PR #8, opened 2020-12-15
feat(address-endpoints): added delete_address

sudoku-lord:
## What

Adds `delete_address` to OpenAPI.


resources/addresses/address.yml
```diff
@@ -20,3 +20,26 @@ get:
...
+  parameters:
+    - in: path
+      name: id
+      required: true
+      schema:
+        $ref: 'attributes/adr_id.yml#/id'
+
+  responses:
+    '200':
+      $ref: responses/address.yml
```
metalsheep, line 45: does the delete actually return?  We have some reason to believe that it doesn't based on working with this in webhooks.

senechko: i dont see this file `responses/address.yml` in the repo or PR

hilary: > i dont see this file `responses/address.yml` in the repo or PR

It's there. The path is relative: https://github.com/lob/hackathon-spec-v1/blob/main/resources/addresses/responses/address.yml

hilary: We don't actually return the address when deleting it (presumably because doing so would make it hard to make the delete endpoint idempotent). The response for delete endpoints is different, returning the id and `deleted: true`. It's the same format for all of our `delete` (ahem, cancel) endpoints, so you might as well put that response in `shared/responses`.

hilary (changes requested):
A bit more complicated than it looked! Almost there.

shared/responses/delete.yml
```diff
@@ -0,0 +1,13 @@
...
+
+content:
+  application/json:
+    schema:
+      type: object
+      properties:
+        id:
+          schema:
+            $ref: ../resources/addresses/attributes/adr_id.yml
+        message:
```
hilary, line 11: Cancel P/D/L/C uses the same response object, so let's put the whole response in `shared/responses` as `deleted.yml`.

The deleted response object needs a little love, too:

```
properties:
  id:
    $ref: ../resources/addresses/attributes/adr_id.yml 
  deleted:
    type: boolean
    required: true
    example: true
```

hilary (changes requested):
getting there...

resources/addresses/address.yml
```diff
@@ -21,3 +21,26 @@ get:
...
+  parameters:
+    - in: path
+      name: id
+      required: true
+      schema:
+        $ref: 'attributes/adr_id.yml#/id'
+
+  responses:
+    '200':
+      $ref: ../shared/responses/delete.yml
```
hilary, line 46: '../../shared/responses/delete.yml'

shared/responses/delete.yml
hilary: If we look at https://docs.lob.com/#addresses_delete it shows the response as

```
{
  "id": "adr_43769b47aed248c2",
  "deleted": true
}
```
The property `deleted` is a `boolean` with value `true`, not a `string` with `value` "deleted: true". 

I realize that this stuff is squirrely. Being able to run prism makes all the difference.

hilary (changes requested):

hilary:
superceeded by #17

PR #9, opened 2020-12-15
feat(shared-headers): created headers file in shared/

sudoku-lord:
## What

Moves the contents of `shared/headers.yml` into `shared/headers/rate-limiters.yml`, and adds `create-address.yml` to `shared/headers/`.


shared/headers/create-address.yml
```diff
@@ -0,0 +1,7 @@
+send_date:
+  description:
+    A timestamp of the date on which a mailpiece will be dispatched
+    for production. Beyond this date, the mailpiece can't be cancelled.
+  schema:
+    type: string
+    example: "2017-09-05T17:47:53.967Z"
```
hilary, line 7: This file is missing a final carriage return. You can configure VS Code to clean your files for you. In Settings => Text Editor => Files, enable Insert File Newline, Trim Final Newlines and Trim Trailing Whitespace and your files will always be lovely. :slightly_smiling_face:

shared/headers/create-address.yml
```diff
@@ -0,0 +1,7 @@
+send_date:
+  description:
+    A timestamp of the date on which a mailpiece will be dispatched
+    for production. Beyond this date, the mailpiece can't be cancelled.
+  schema:
+    type: string
```
hilary, line 6: We can use `format: date-time` here as well as string. https://swagger.io/docs/specification/data-models/data-types/#string

shared/headers/create-address.yml
```diff
@@ -0,0 +1,7 @@
+send_date:
+  description:
```
hilary, line 2: W/re [YAML block scalars](https://yaml.org/spec/1.2/spec.html#id2793652), I have found that being explicit saves so much later headache. Unless our source material indicates otherwise, all our block scalars are folded `>`, and we should assume that we are stripping any trailing lines, as well. `-`. (And yes, I have some cleanup to do from earlier work w/re stripping because I was working from Luke's scrape rather than our hbs templates.)

hilary (changes requested):
Let's not call this header `create_address.yml`, let's call it `send_date.yml`, since it is used by multiple types of resources.

hilary:
Ok, so this is totally my mistake. I misunderstood the documentation. send date is a shared attribute, not a shared header. What would you like to do? We still need to shift rate limiters to a shared headers folder because we also have the idempotency key header. I'm tempted to suggest submitting just those changes for now.

hilary:
Replaced by #16

PR #10, opened 2020-12-15
Spectral cleanup

hilary:
A tiny PR with some Spectral improvements that arose as a side-effect of some debugging I'm doing.

* I removed a rule `operation-singular-tag: off` because it is no longer needed; `operation-singular-tag` is now set to `off` by default.
* I enabled the full Spectral OAS v3 ruleset (`extends: [[spectral:oas, all]]`). Using `extends: spectral:oas`, which is what we had before, enables a subset of the Spectral OAS v3 ruleset.

### DIFF
--- a/.spectral.yml
+++ b/.spectral.yml
@@ -1,7 +1,6 @@
 extends: spectral:oas
 rules:
   operation-default-response: off
-  operation-singular-tag: off
 
   path-item-must-be-ref:
     description: Path Item Object value must be a $ref
--- a/.spectral.yml
+++ b/.spectral.yml
@@ -1,4 +1,4 @@
-extends: spectral:oas
+extends: [[spectral:oas, all]]
 rules:
   operation-default-response: off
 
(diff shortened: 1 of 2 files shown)


rkofman-lob:
Could you explain what the cleanup actually does/affects?

rkofman-lob (approved):
Seems legit. Would like the PR description to explain the change a teeny bit more (so code archeologists can know why we thought `operation-singular-tag` was useful, and how we decided it wasn't.

hilary:
> Seems legit. Would like the PR description to explain the change a teeny bit more (so code archeologists can know why we thought `operation-singular-tag` was useful, and how we decided it wasn't).

Sorry about that - I have those explanations in the individual commit messages since commits are forever, whereas PRs are github gloss. (Code archeologists will see commits but may or may not actually see PRs.) It's my bad as I am a fan of surfacing info in PRs! I'll make sure to surface that info in PR messages as well!

rkofman-lob:
Makes sense (and agreed on importance of commits > PR). I'm just used to a workflow where commit messages often got squashed and discarded rather than kept in any useful form.

PR #11, opened 2020-12-16
add error responses

hilary:
This PR adds an [error model](https://lobsters.atlassian.net/browse/HAS-27) that is shared by all endpoints and uses that error model with the default response to model(*) the failure responses for the two existing endpoints (https://lobsters.atlassian.net/browse/HAS-8).

(*) See "Default Response" on https://swagger.io/docs/specification/describing-responses/

### DIFF
--- a/.spectral.yml
+++ b/.spectral.yml
@@ -1,6 +1,5 @@
 extends: [[spectral:oas, all]]
 rules:
-  operation-default-response: off
 
   path-item-must-be-ref:
     description: Path Item Object value must be a $ref
--- a/shared/attributes/failure_status_code.yml
+++ b/shared/attributes/failure_status_code.yml
@@ -1,0 +1,20 @@
+type: integer
+
+enum:
+  - 401
+  - 403
+  - 404
+  - 422
+  - 429
+  - 500
+
+description: >
+  A conventional HTTP status code:
+    * 401 - Authorization error with your API key or account
+    * 403 - Forbidden error with your API key or account
+    * 404 - The requested item does not exist
+    * 422 - The query or body parameters did not pass validation
+    * 429 - Too many requests have been sent with an API key in a given amount of time
+    * 500 - An internal server error occurred, please contact support@lob.com
+
+example: 429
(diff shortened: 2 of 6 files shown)


shared/attributes/failure_status_code.yml
```diff
@@ -0,0 +1,20 @@
+type: integer
+
+enum:
```
rkofman-lob, line 3: Have to watch out for enum values / make sure any tooling we use for auto-generation is intelligent. If it translates to a formal Enum in Java, and we decide to add another error code, an SDK can blow up. (ask me how I know)

Not blocking here; but wanted to add context in case we're not yet thinking about this.

.spectral.yml
```diff
@@ -1,6 +1,5 @@
 extends: [[spectral:oas, all]]
 rules:
-  operation-default-response: off
```
rkofman-lob, line 3: curious why this used to be a lint rule?

shared/attributes/failure_status_code.yml
```diff
@@ -0,0 +1,20 @@
...
+  - 429
+  - 500
+
+description: >
+  A conventional HTTP status code:
+    * 401 - Authorization error with your API key or account
+    * 403 - Forbidden error with your API key or account
+    * 404 - The requested item does not exist
+    * 422 - The query or body parameters did not pass validation
+    * 429 - Too many requests have been sent with an API key in a given amount of time
```
rkofman-lob, line 17: In the future, might be useful to link out to any formal guidance we provide publicly about rate limits.

rkofman-lob (approved):

.spectral.yml
hilary: So that linting would pass while we were bootstrapping the spec to this point.

shared/attributes/failure_status_code.yml
hilary: Good thought! I am a huge fan of contextual documentation. @nihwang will be looking at the spec from the docs point of view starting in Q1. Letting her know about your suggestion.

.spectral.yml
rkofman-lob: Oh; I see. This was turning *off* linting, and now it's back on. I misread it as linting for not having default responses -- which now that I think about it makes a lot less sense.

PR #12, opened 2020-12-16
update spec to match reality

hilary:
While the documentation (and therefore the spec) specify a two letter
(ISO 3166) country code, the value returned for the example in the docs
does not conform. The discrepancy reveals that the field does not have
adequate restrictions on the backend. I removed the restriction on the
spec (although I did not change the wording) until we can fix the
backend.

### DIFF
--- a/resources/addresses/models/address.yml
+++ b/resources/addresses/models/address.yml
@@ -112,7 +112,6 @@ properties:
     description: >
       Must be a 2 letter country short-name code (ISO 3166). Defaults to <code>US</code>.
     minLength: 2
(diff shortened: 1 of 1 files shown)


rkofman-lob (approved):
eeek. Is there a Jira for follow-up backend work (and/or a plan to update the text to represent reality)?

hilary:
> eeek. Is there a Jira for follow-up backend work (and/or a plan to update the text to represent reality)?

I notified @jeff-wilde via Slack. I can create a JIRA if desired.

PR #15, opened 2021-01-05
add 413 to failure status codes

hilary:
What it says :slightly_smiling_face: 

No, seriously. I realized, during the Atlas standup this morning, that our failure list was missing [413](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml), which Dora may return.


lbirdeau (approved):
LGTM

PR #19, opened 2021-01-06
move metadata to models

hilary:
objects are models, not attributes (duh)
also removed redundant model name


shared/models/metadata.yml
```diff
@@ -0,0 +1,12 @@
+type: object
+description: When creating any Lob object, you can include a metadata object with up to 20 key-value pairs of custom data.
+properties: {}
```
lbirdeau, line 3: Not sure how you feel about explicitly defining the `additionalProperties` field for objects like this, but I think it could be helpful here. This particular field seems pretty nuanced and open to error. For example, it it really is possible to passing metadata fields where the value can be a number, string, or boolean, then we may want to document these possibilities in the example and then do something like
```
additionalProperties: true
```

But we might find that due to legacy issues with our current customer base that there are a range of customers that do not submit JSON (and instead url-encode their submissions), and therefore, the metadata field might not actually support data typing of any sort (we cannot resolve anything except for strings). In this case we can explicitly let customers know of this limit by keeping the example as is but adding:
```
additionalProperties:
  type: string
```

hilary: I don't believe the url-encoding is a complication in this case. I do think we could do a better job of defining this field, though!

I need to make sure I understand exactly what the legal options for the metadata object are, first.

hilary: I created a ticket to flesh out the metadata definition: https://lobsters.atlassian.net/browse/HAS-41
This PR is just to move metadata from attributes to models, since objects should be in models, not attributes.

lbirdeau (approved):
LGTM

PR #21, opened 2021-01-06
remove redundant model name

hilary:
for consistency!

### DIFF
--- a/shared/models/list.yml
+++ b/shared/models/list.yml
@@ -1,25 +1,24 @@
-List:
-  type: object
-  description: Multiple items returned in order
-  properties:
-    object:
-      type: string
-      description: What kind of resource does this list contain?
-      example: address
+type: object
+description: Multiple items returned in order
+properties:
+  object:
+    type: string
+    description: What kind of resource does this list contain?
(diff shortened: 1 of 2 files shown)


shared/models/list.yml
```diff
@@ -1,25 +1,24 @@
...
-    previous_url:
-      type: string
-      description: Url of previous page of items in list.
-      nullable: true
-      example: null
+  previous_url:
+    type: string
+    description: Url of previous page of items in list.
+    nullable: true
+    example: null
```
lbirdeau, line 19: An example URL would be nice here (even though the url structure is arguably opaque).

lbirdeau (approved):
LGTM

shared/models/list.yml
hilary: It isn't so much that the structure is arguably opaque as that there's no simple way to override an incorrect example, at least in OpenAPI v3.0. A lot of work went into examples in the upgrade between 3.0 and 3.1.

We'll need to wait until we've either got an extension in place or are on 3.1 to have examples for this one, unfortunately.

PR #25, opened 2021-01-15
add contract tests

hilary:
https://lobsters.atlassian.net/browse/ENG-15299

This PR contains a new contract testing framework and the contract tests for the address endpoints.

### DIFF
--- a/tests/setup.js
+++ b/tests/setup.js
@@ -1,0 +1,26 @@
+'use strict';
+
+const { getHttpOperationsFromSpec }  = require('@stoplight/prism-cli/dist/operations');
+const { createClientFromOperations } = require('@stoplight/prism-http/dist/client');
+const { URL }                        = require('url');
+
+const lobUri = 'https://api.lob.com/v1';
+const clientOptions = {
+  mock: false,
+  validateRequest: true,
+  validateResponse: true,
+  checkSecurity: false,
+  errors: true,
+  baseurl: lobUri,
+  upstream: new URL(lobUri),
+};
+
+const specFile = './Lob-API-public.yml';
+
+module.exports.setup = async () => {
+  const operations = await getHttpOperationsFromSpec(specFile);
+  return createClientFromOperations(operations, clientOptions);
+}
+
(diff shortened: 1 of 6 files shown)


tests/setup.js
```diff
@@ -0,0 +1,26 @@
...
+};
+
+const specFile = './Lob-API-public.yml';
+
+module.exports.setup = async () => {
+  const operations = await getHttpOperationsFromSpec(specFile);
+  return createClientFromOperations(operations, clientOptions);
+}
+
+const testSecret = 'BASIC dGVzdF8wZGM4ZDUxZTBhY2ZmY2IxODgwZTBmMTljNzliMmY1YjBjYzo='
```
lbirdeau, line 25: I think there might be a linter command that looks for keys like this.

lbirdeau (changes requested):
Key

tests/setup.js
lbirdeau: nvr mind. looks like this particular endpoint is meant to be public for now. I'll approve.

lbirdeau (approved):

tests/setup.js
hilary: It's the key from the developer docs site, so not adding any additional badness. It's also temporary until we do the ticket that pulls the token from the environment and does the base64 encoding inline. Which is all in the README. :smile:

hilary: Nonetheless, I super appreciate your careful reading and concern for security!

tlhunter, line 25: I've never seen basic auth using a spelling other than 'Basic'! I've definitely wrote implementations that check for that exact string. Makes me think of the HtTp header casing.

tlhunter, line 25: I've never seen basic auth using a spelling other than 'Basic'! I've definitely wrote implementations that check for that exact string. Makes me think of the HtTp header casing.

tlhunter (commented):
LGTM

tests/setup.js
hilary: :woman_facepalming: what an excellent point!

PR #26, opened 2021-01-19
check response.status in tests, fix sPeLling of BaSIC

hilary:
promises are tricksy

* We have to resolve the promise in order to determine whether the response was actually ok. Tests updated to check response.status. Leaving aside that these shouldn't be returning 200...
* Fix for @tlhunter's #25


lbirdeau (approved):
LGTM

PR #27, opened 2021-01-20
add zip_lookups

hilary:
* adds endpoints for zip_lookups (https://lobsters.atlassian.net/browse/ENG-15317)
* includes a couple of refinements to the contract testing framework:
  * we now have `tape-catch` in our devDependencies; I found it invaluable in debugging the test issue I was chasing down, even though I didn't keep it in the final version of the test
  * `setup` now accepts an `override` object for per test client settings (used here to set `errors: false`)

I ran into the errors + promises + testing maze with one of these tests. I ended up rewriting the test to check the returned error message rather than checking the thrown error. I couldn't get `t.throws` to accept an async function for its first parameter regardless of how I wrote it (async/await, promise, etc., etc.). The documentation for `t.throws` refers you to the documentation to Node.js [`assert.throws`](https://nodejs.org/api/assert.html#assert_assert_throws_fn_error_message) for examples; based on Node's docs, I'm not sure that it can take a function that returns a promise. The current pattern should work fine for our needs.

### DIFF
--- a/.spectral.yml
+++ b/.spectral.yml
@@ -20,4 +20,4 @@ rules:
     then:
       function: pattern
       functionOptions:
-        match: "^resources\/[a-zA-Z]+\/[a-zA-Z-_]+.yml"
+        match: "^resources\/[a-zA-Z-_]+\/[a-zA-Z-_]+.yml"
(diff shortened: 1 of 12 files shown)


lbirdeau (approved):
LGTM

PR #30, opened 2021-01-28
task(endpoint-creation): added verify endpoint

sudoku-lord:


sudoku-lord:
The issue I see here is that the `address` field shouldn't be lumped in with the `primary_line`, `city`, `state`, etc. There are TWO ways to configure the verification payload––either with multiple fields representing address components, _or_ a single-line address.

hilary (changes requested):
You're right about not lumping `address` in with the other options. And you're right about it being related to the extension.

We're in a funky situation which derives from writing the spec after the fact rather than before. With Prism, we can write contract tests that go beyond what the spec actually expresses. That's a good thing as far as I'm concerned; it's one of the strengths of the tool.

What's more, it applies beyond our current circumstance, because it means that we can be flexible as we move forward. We're not restricted to what we can currently express in OpenAPI vWhatever.

To solve:
1. Use separate models for input and response
2. Move common attributes to `shared/attributes`
3. For the input model, use `oneOf`

sudoku-lord:
Alright, so it looks _mostly_ okay, but there are two things I'm unsure of. I can toggle between request forms (single address vs multi-component), but they're both labelled "object". Should I be defining some kind of separate type or name for them?

Skjermbilde 2021-02-01 kl  6 42 05 pm

Also, should we show a view for each potential request configuration? Right now, only single-line is showing up and I'm not sure whether we should be toggling between them or something.

resources/us_verifications/models/components.yml
```diff
@@ -0,0 +1,232 @@
...
+  extra_secondary_number:
+    type: string
+    description: >
+      An extra (often unnecessary) secondary number provided with the input address.
+    example: ""
+    readOnly: true
+
+  city:
+    readOnly: true
+    allOf:
```
hilary, line 115: So even though the linters don't flag these, they aren't conforming to the standard:

" allOf takes an array of object definitions that are validated independently but together compose a single object."

Go ahead and leave out the `ReadOnly`s, as the behavior of those change in v3.1.

hilary (changes requested):
Fix the `allOf`s that are nonconforming (use `allOf` with an attribute, not an object) and we're good!

hilary (approved):
ship it!

PR #31, opened 2021-01-29
Push bundled spec to repo

hilary:
Including the bundled spec as one of our release artifacts is a goal in and of itself, although I didn't think to give it its own ticket. :woman_facepalming:

Having the bundled spec as part of the github action workflow is also necessary for producing the postman collection, which is https://lobsters.atlassian.net/browse/ENG-15321.

In other news, figuring out this workflow was crazy pants, and I would like very much to spend a day writing up the whole thing!

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,0 +1,27 @@
+name: CD
+
+on: workflow_dispatch
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


senechko (approved):

.github/workflows/cd.yml
```diff
@@ -0,0 +1,27 @@
+name: CD
```
lbirdeau, line 1: I'm going to use this for the address parser API spec as well. Thx for figuring this out.

lbirdeau (approved):
LGTM

PR #32, opened 2021-02-03
push changes only when there are changes

hilary:
Fix a problem with #31 
- when merging to main, the branch already has the bundled spec: the subsequent push fails because there are no changes
- in any case, no need to push when the bundled spec hasn't changed! Now checking to see if it has actually changed before staging, committing and pushing
- using `git status --porcelain` instead of plumbing commands because the check reports as a failure if any step does something which returns a non-zero exit code. Note: it might be possible to redirect the exit code to the void and survive, but that defeats the purpose of using a git plumbing command. Anyhoo...

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -4,7 +4,7 @@ on: workflow_dispatch
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
 
     steps:
       - name: checkout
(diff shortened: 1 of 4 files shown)


.github/workflows/cd.yml
```diff
@@ -4,7 +4,7 @@ on: workflow_dispatch
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
lbirdeau, line 7: why the change to 'latest'?

lbirdeau (approved):
LGTM

.github/workflows/cd.yml
hilary: The specific `ubuntu-20.04` tag turns out to be listed as `experimental`, whatever that means.

PR #33, opened 2021-02-08
pull test token from environment

hilary:
A necessary step along the path to building out our contract tests in github actions.

PR #34, opened 2021-02-08
make contract tests more configurable

hilary:
I've started thinking of `setup.js` as `tape-prism` if anyone is wondering what the commit message is about. 

This PR comes after #33 . As a github action, we want the test runner to take configuration from outside the runner itself, so this step does just that: it takes the spec file, etc., as configuration options.

I'm pushing ahead to get through this set of changes. I expect that I'll come back and refactor once I've got everything done. That said, I'm more than happy to get specific refactoring suggestions along the way.


tests/setup.js
```diff
@@ -3,24 +3,36 @@
 const { getHttpOperationsFromSpec }  = require('@stoplight/prism-cli/dist/operations');
 const { createClientFromOperations } = require('@stoplight/prism-http/dist/client');
 const { URL }                        = require('url');
+const btoa                           = require('btoa');
```
senechko, line 6: @hilary can you add prettier to this repo and be done with this assignment/alignment stuff?

hilary: Yup, will do.

senechko (approved):

PR #36, opened 2021-02-09
run contract tests in CI

hilary:
Run our contract tests in CI: https://lobsters.atlassian.net/browse/ENG-15379

I built this out as a private github action for now (it will live in each repo - see the `README.md` in `/actions/contract_tests` for more info). We can extract it out into a public github action at a later time.


lbirdeau (approved):
LGTM

PR #42, opened 2021-02-17
improve prettier setup

hilary:
Fixes some issues we had with our prettier setup:
- sometimes prettier would try to reinitialize and add `lint-staged` (which adds `mrm`)
- the pre-commit githook wasn't running
- `pretty-quick` has `master` hard-coded as the default branch for github

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,1 @@
+printWidth: 80
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npm test && npx prettier --check .
--- a/.prettierignore
+++ b/.prettierignore
@@ -6,3 +6,8 @@ node_modules
 *.html
 *.hbs
 
+# File exists to prevent prettier from re-initializing
+# prettier and introducing lint-staged. We don't want
+# lint-staged because it introduces a metric ton of
+# dependencies and generally takes over the universe.
+.prettierrc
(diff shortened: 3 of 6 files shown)


senechko (approved):

.prettierrc
```diff
@@ -0,0 +1 @@
+printWidth: 80
```
senechko, line 1: 80 is the default so i dont think this is necessary 
image

https://prettier.io/docs/en/options.html

hilary: I found that prettier seemed to ignore the presence of an empty file. :disappointed:

rogue0137 (commented):
@hilary was the main reason for removing `pretty-quick` to get rid of it's reliance on using `master` as the default branch?

hilary:
> @hilary was the main reason for removing `pretty-quick` to get rid of it's reliance on using `master` as the default branch?

The recommended design adding the `husky`/`pretty-quick` `JSON` stanza in `package.json` wasn't working. After some debugging, I noticed that `master` was built-in to `pretty-quick`. I did experiment `pretty-quick`s `--branch` option, but that didn't work either. At that point I looked at husky's docs again, saw that husky didn't have examples with stanzas in `package.json`, but used the traditional githook setup, breathed a sigh of relief and switched.

Looking today, it appears as if husky v4 supported stanzas in `package.json`, while husky v5 does not / no longer needs them / whatever. My normal inclination would be to research this until I understood every nuance, but I'm trying to stay on task. :stuck_out_tongue_closed_eyes:

rogue0137 (commented):
testing

PR #45, opened 2021-03-02
fix(required): added XOR required statements in necessary places

sudoku-lord:
## Background

Some properties in models like `address_writable` and `multiple_components` (used for creating verifications) are required only when certain other properties are not present. This PR uses the `oneOf` keyword to fix that.

After much researching, @hilary and I determined that this is in fact legal in OpenAPI 3.0.3 (and OpenAPI 3.1.0). 

## Note

`address_writable` has some extra weirdness related to the `address_country` which I will address in a subsequent PR (by creating separate models for international and US).


resources/addresses/models/address_writable.yml
```diff
@@ -3,6 +3,12 @@ type: object
 required:
   - address_line1
 
+oneOf:
```
hilary, line 6: This should be `anyOf`. "Either name or company is required, you may also add both." https://docs.lob.com/#addresses_create

hilary (changes requested):
We need tests for these various cases, given how tricky encoding them is.

hilary (approved):
This PR is my fave!

:shipit: !!

PR #47, opened 2021-03-04
fix spectral lint failure

hilary:
quick lint failure fix to unblock other work


dist/Lob-API-bundled.yml
```diff
@@ -526,7 +526,7 @@ components:
           description: >
             Geographic direction preceding a street name (`N`, `S`, `E`, `W`,
             `NE`, `SW`, `SE`, `NW`).
-          example: ''
+          example: 'N'
```
sudoku-lord, line 529: I believe this example will show up in the example address given (the 185 BERRY ST one), which doesn't actually have any directionals IRL. Should we change the address to one which does have pre- and post-directionals, which would be a break from the current docs, or is it okay to put in a directional where there isn't one?

dist/Lob-API-bundled.yml
```diff
@@ -555,7 +555,7 @@ components:
           description: >
             Geographic direction following a street name (`N`, `S`, `E`, `W`,
             `NE`, `SW`, `SE`, `NW`).
-          example: ''
+          example: 'N'
```
sudoku-lord, line 558: Same question as the one above.

resources/us_verifications/models/components.yml
```diff
@@ -51,7 +51,7 @@ properties:
     description: >
       Geographic direction preceding a street name (`N`, `S`, `E`,
       `W`, `NE`, `SW`, `SE`, `NW`).
-    example: ""
+    example: "N"
```
sudoku-lord, line 54: Same question as above, and ditto for the next one.

dist/Lob-API-bundled.yml
hilary: Ok, so I looked at the documentation again, and it appears to me as if what we need is to include the empty string in the enum list, as both pre and post directionals are required.

dist/Lob-API-bundled.yml
hilary: Same answer as above :)

sudoku-lord (approved):

PR #48, opened 2021-03-08
Add templates, template versions resources

hilary:
https://lobsters.atlassian.net/browse/ENG-15308

prioritized to unblock https://lobsters.atlassian.net/browse/PMAPI-282

Note that we're moving examples away from individual properties and into requests and responses to be forward compatible with 3.1. I'll submit a separate PR that makes that conversion for the rest of the existing resources.


resources/templates/template_versions/models/template_version.yml
```diff
@@ -0,0 +1,26 @@
...
+
+  - type: object
+
+    required:
+      - id
+      - date_created
+      - date_modified
+      - object
+
+    properties:
```
zacleids, line 12: a new property called `engine` was added to template versions. its currently `null` but will be an enum of 'legacy' and 'handlebars'

zacleids: its also missing a boolean called `suggest_json_editor`

hilary: - @zacleids the `description` is inherited from `template_version_writable`.
- w/re `suggest_json_editor`, 
  - could I please get a description of the property (what does it mean)?
  - does it have a default value?
  - is it writable, i.e., can it be set via the data submitted in a `POST` create request) or is it only set by the backend?
- w/re `engine`
  - could I please get a description of the property (what does it mean)?
  - the property is a string enum?

hilary: added `suggest_json_editor` and `engine`. We can flesh out specifics when they become available.

zacleids: sorry for the late reply. suggest_json_editor is a flag used by the frontend to see if we should suggest the json editor. It is true if the template is using advanced features.

The engine field is an enum with 'legacy' and 'handlebars' and determines which engine to use to combine their html template with their merge variables

hilary: no worries @zacleids , updated with your info

sudoku-lord (approved):

PR #50, opened 2021-03-19
add tracking events

hilary:
This PR has work common to postcards and letters extracted into an independent PR. @sudoku-lord is working on postcards in parallel with my work on letters. This PR is the first of several from one or the other of us arising from coordinating efforts and avoiding duplication. 

Because the changes in the PR are isolated, they do not contain tests for the tracking event models. As we don't have a documented endpoint for customers to fetch tracking events by id, these models are tested indirectly, via the models in which they are embedded. Not optimal. :crying_cat_face: 

This PR also necessarily contains a refactor that predates adding the tracking events. Ideally, I would postpone that refactor or submit it separately. Resource and time constraints being what they are, I have included it.


senechko (approved):

sudoku-lord (commented):
Looks good!

PR #53, opened 2021-03-22
created address_lines

sudoku-lord:
## What

Created PRs to separate out address lines models

## Why

They're used eeeeverywhere.


hilary (commented):
I think we're ready: take a look. We probably want to update the README to reflect what we're doing with shared. We should also add info about the generator to the README, but can do both in a separate PR.

hilary (approved):

PR #54, opened 2021-03-24
connect tracking events

hilary:
- connect tracking events to letters
- remove some old files that got missed in previous PRs (doh!)
- refactor `address_editable` model to DRY it a bit
- remove `example` from individual attributes (and `readOnly` when I noticed it since it will go away in v3.1)


sudoku-lord (approved):

sudoku-lord (changes requested):
It looks like `models/letter.yml` is missing the `url` field which is described in the [docs](https://docs.lob.com/#letters_object).

hilary:
> It looks like `models/letter.yml` is missing the `url` field which is described in the [docs](https://docs.lob.com/#letters_object). The URL for the custom envelope field is present, but there's also a URL associated with the letter object (which is how the customer downloads it).

nice catch!

sudoku-lord (approved):
I'm wondering whether the examples in the response files should contain every field, or only required ones? I saw that in the examples within the responses for `letter` and `all_letters`, not every field is listed out.

hilary:
> I'm wondering whether the examples in the response files should contain every field, or only required ones? I saw that in the examples within the responses for `letter` and `all_letters`, not every field is listed out.



> I'm wondering whether the examples in the response files should contain every field, or only required ones? I saw that in the examples within the responses for `letter` and `all_letters`, not every field is listed out.

We should work towards having multiple examples. At least one that contains every field, and at least one that has only the required fields. For now, we're doing a first pass through the documented endpoints. The examples aren't going to be ideal.

Please add a section (maybe even a page?) to the markdown docs discussing examples:
- where to put them (in requests/responses and models)
- why (3.1 compatibility)
- best practices (multiple examples, at least one with all, at least one with only required, cover cases of interest)
- use cases: link to prism mock docs on using named examples

PR #56, opened 2021-03-25
add test for certified letters

hilary:
Now that we have the lob-openapi account setup to be able to send certified, etc., in test mode, I can reason more fluently about how to model the polymorphism in the various resources. I refactored some more, creating cleaner experiences for developers consuming the API.


senechko (approved):

resources/letters/models/letter_base.yml
```diff
@@ -34,6 +35,7 @@ properties:
     oneOf:
       - $ref: "../../../shared/attributes/model_ids/adr_id.yml"
       - $ref: "../../../shared/models/address/inline_address.yml"
+      - $ref: "../../../shared/models/address/address_editable.yml"
```
sudoku-lord, line 38: Why is this either inline address or address editable?

sudoku-lord (approved):

resources/letters/models/letter_base.yml
hilary: Good catch. Feel into that encoding trap yet again. I needed to move `to` and `from` out of `base` and into `editable` and `generated_base`.

`to` and `from` in `generated_base` are now the full address model, which I rather thought they should be. All of the address model is now shared.

PR #58, opened 2021-03-26
Add certificate endpoint

imWillX:
Adds certificate endpoint and tests


resources/certificates/responses/all_certificates.yml
```diff
@@ -0,0 +1,18 @@
...
+      properties:
+        data:
+          type: array
+          description: list of certificates
+          items:
+            $ref: "../models/certificate.yml"
+        count:
+          $ref: "../../../shared/attributes/count.yml"
+        object:
+          $ref: "../../../shared/attributes/object.yml"
```
hilary, line 18: At some point, we'll want to add an example to this file

resources/certificates/responses/certificate.yml
```diff
@@ -0,0 +1,6 @@
+description: Returns an certificate object if a valid identifier was provided.
+
+content:
+  application/json:
+    schema:
+      $ref: "../models/certificate.yml"
```
hilary, line 6: At some point, we'll want to add an example to this file as well.

tests/certificate_test.js
```diff
@@ -0,0 +1,69 @@
...
+
+  // delete
+  response = await prism
+    .setup()
+    .then((client) =>
+      client.delete(tmpl_endpoint, { headers: prism.authHeader })
+    );
+
+  t.equal(response.status, 200);
+});
```
hilary, line 69: At some point, we should update the tests to the newer design we're following (see letters or the generator for examples.)

hilary (approved):
I noted some enhancements we can make, but I don't think we need them before merging.

Looking great!

PR #62, opened 2021-04-12
[WIP] extract form factor base models [DXP-112]

hilary:
I'm still getting the address models to work the way we need them to in all their glory, but the form factor and lob base work is done.

The core of this PR is:

- `lob_base.yml` All non-AV Lob resources (*) have these properties, whether Form Factors or not
- `shared/models/form_factor`  Properties common to all Form Factors
  -  `editable.yml` Properties which customers can customize
  - `generated.yml` Properties which we generate
  - `to.yml` The `to` address, which has rules sufficiently complex that it is its own model. Thus far, `to` is the same across Form Factors.
  - `from.yml` The `from` address has different rules sufficiently complex that it is own model.
  - `from_us.yml` Most Form Factors only allow `from` addresses in the US

And then there is:
- `shared/models/address` Address models, US and International, editable and stored, inline vs not. The address models are driven by a combination of how polymorphism and inheritance work in OpenAPI v3 and the fact that we require a 2-digit country code in the input model which we then convert into the full length name for the printed address. Because the possible values for the discriminator for a polymorphic class must be enumerated, the full list of country codes has to be listed in `shared/models/address/address.yml` and `shared/models/address/address_editable.yml`.

(*) Certificates do not :sob:


hilary:
replaced by #71

PR #63, opened 2021-04-12
task(files): add local filepath for letters

sudoku-lord:
## What

Adds options for remote & local filepaths for letters. Also does a little shifting around of attributes.


dist/Lob-API-bundled.yml
```diff
@@ -1062,6 +1062,10 @@ components:
       oneOf:
         - $ref: '#/components/schemas/inline_address_us'
         - $ref: '#/components/schemas/inline_address_intl'
+    local_file_path:
+      type: string
+      description: A link to a remote file to be used for the postcard template.
+      pattern: '[a-zA-Z0-9@:%._+~#=/]{1,256}.(html|pdf)+$'
```
senechko, line 1068: is this really the constraint? like no `.htm` files?

sudoku-lord: Hmm, the docs say HTML files. It didn't strike me to include `htm` but I don't see why it's unacceptable... I can try and upload an HTM file in staging to check.

sudoku-lord: Just checked and the docs do word it as "a local upload of an HTML or PDF file" with no mention of HTM files. What's even weirder is that when I download an HTML template STRAIGHT from [the gallery linked in the docs](https://www.lob.com/template-gallery) and then navigate to the Postcards dashboard to create a test postcard, I'm unable to upload the template, regardless of whether its suffix is `html` or `htm`. Is this a bug of some kind? @hilary

hilary: Does `htm` work via `curl`?

sudoku-lord: Filed a bug ticket with Print & Mail API.

sudoku-lord: Addendum: HTML & HTM both work with `curl`, so I'll add HTM to the list of extensions.

tests/postcards_test.js
```diff
@@ -95,7 +95,7 @@ test("create, list, read then cancel a postcard", async function (t) {
   await deleteAddress(from);
 });
 
-test("creates a postcard given a local filepath as the front", async function (t) {
+test.only("creates a postcard given a local filepath as the front", async function (t) {
```
hilary, line 98: definitely don't want to leave this in place!

resources/letters/attributes/local_file_path.yml
```diff
@@ -0,0 +1,6 @@
+type: string
+
+description: >-
+  A link to a remote file to be used for the postcard template.
```
hilary, line 4: Lets put `local_file_path.yml` in `shared/attributes` and add a comment block that shows how to use it, with `allOf` and the `pattern: "ridiculouslycomplexregex"`.

The description can be `A link to a remote file to be used for the template.`

The goal is to highlight the similarities and differences between form factors....

hilary (changes requested):

tests/postcards_test.js
sudoku-lord: OOPS, getting rid of it now!

resources/letters/attributes/local_file_path.yml
sudoku-lord: Ah, so the reason there are two `local_file_path` files is that they are slightly different between letter & postcard (letter does not allow PNG or JPEG uploads).

hilary (approved):
LGTM

PR #64, opened 2021-04-13
fix(list): added parameters to addresses list functionality

sudoku-lord:
## What

The `addresses.yml` file, which specs out the list operation for the `address` resource, previously did not contain the full range of listing parameters specified by the docs. As far as I could see, this is the only place where they're missing, so I added them in. 

Modifying the tests was mentioned, but I'm not sure that's strictly necessary. This is what a sample test (from `letters_test.js`––`letters.yml` already contained the full range of parameters) looks like:

```
  const list = await prism
    .setup()
    .then((client) =>
      client.get(resource_endpoint, { headers: prism.authHeader })
    );
  t.equal(list.status, 200);
```
There are no references to the parameters here––it looks pretty much identical to how the listing test from `addresses_test.js` looked from the beginning.


hilary (approved):
:shipit:

PR #69, opened 2021-04-20
PMAPI-282: add template compile route, migrate make scripts to npm scripts, add documentation

zacleids:


hilary:
I dealt with the vulnerability. The library in question is a transitive dependency that is used by a dev script. The issue has already been fixed in a later version of the library, and I've observed that the invoking library (redoc-cli) clears issues identified by npm audit reasonably quickly so went ahead and marked it "won't fix".

hilary (approved):
Terrific job @zacleids !

Thanks so much for all the improvements!!

PR #73, opened 2021-04-20
fix(template): fixed minor issue in config

sudoku-lord:
aaaa


sudoku-lord:
oh it was fixed already lol

PR #74, opened 2021-04-22
feat(contract-tests): add script which allows for running tests by goalie

sudoku-lord:


hilary (approved):
Looks great to me :100: 

Would be even greater if you renamed `certificate_test` to `certificates_test`, but not going to hold up approval for that!

PR #78, opened 2021-04-24
streamline workflows

hilary:
This PR contains a set of small changes that stabilize and speed up the CD components of our CI/CD pipeline:

* update the existing contract test Dockerfile to node v14.16 to match the engines stanza now in package.json
* remove the obsolete step that published the zero-dependency version of the developer docs to the dedicated gh-pages branch (I switched our source to `docs/index.html` on main at the beginning of the week, but held off removing the step until I was confident about the build for ease of reverting.)
* leverage the contract tests action to run the postman collection action rather than use an external action; it's faster and works
* add READMEs to both `dist/` and `docs/` directories. While the same information is available elsewhere in the docs, contextual docs are a dev's best friend
* move multi-tape to devDependencies, where it belongs


sudoku-lord (approved):

.github/workflows/cd.yml
```diff
@@ -21,24 +21,25 @@ jobs:
...
+          set -x
           git update-index -q --refresh
           changes=$(git status --porcelain)
           if [[ "${changes}" ]]; then
             git config user.name lobot
             git config user.email lobot@lob.com
+            git add --force dist/Lob-API-postman.json
             git add --force dist/Lob-API-bundled.yml
             git add --force docs/index.html
             git commit -m "update bundled spec, zero-dependency docs"
```
rkofman, line 43: Interesting that we're doing this generation after push and not in e.g. a pre-commit hook that e.g. gets verified as matching the build on push. One the one hand, this seems like it will pollute the git log; on the other hand -- it does rather neatly solve any issues with potential conflicts in the generated files being merged.

rkofman (commented):
:shipit:

.github/workflows/cd.yml
hilary: Well, it isn't perfect by any means. Ultimately, I don't mind the extra commits as they fade into the background pretty much instantly. Hmmm... I think I'll update the commit message to say "CD: update bundled spec, zero-dep docs, postman collection", which should fade even more effectively. As a trade-off for totally reliable updates, felt like a no-brainer to me.

PR #79, opened 2021-04-26
fix(per-goalie): runs tests for all command-line args

sudoku-lord:
## What

Allows you to input more than one team when running the tests with the `pergoalie` script.

Previous: `npm run pergoalie -- av`
Now: `npm run pergoalie -- av pmapi billing`

## Why

Innit cool?


hilary (approved):
Sweet!

PR #83, opened 2021-04-28
added rejection checks for each test

sudoku-lord:
lol the branch name's outdated, we have moved on to `tape-promise`. 

This is one of a set of two PRs in this vein. This one just adds `doesNotReject` checks after every async call in the tests. The next one will deal with the async functions within tests, such as in `letters_test.js`, and might be a little more complex.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
       "staging": "https://api.lob-staging.com/v1",
       "production": "https://api.lob.com/v1"
     },
-    "goaliemappings": {
+    "goalieMappings": {
       "av": {
         "slackchannel": "#av-api-alerts",
         "resources": {
@@ -26,7 +26,7 @@
         "slackchannel": "#billing-errors",
         "resources": {}
       },
-      "pmapi": {
+      "pmApi": {
         "slackchannel": "#eng-api-errors",
         "resources": {
           "addresses": "tests/addresses_test.js",
@@ -79,6 +79,7 @@
     "redoc-cli": "^0.11.4",
     "tap-spec": "^5.0.0",
     "tape": "^5.2.2",
+    "tape-promise": "^4.0.0",
     "uuid": "^8.3.2"
   },
   "engines": {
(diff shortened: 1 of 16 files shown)


hilary (commented):
Let's make sure we fully understand how error messages percolate up before proceeding...

tests/postcards_test.js
```diff
@@ -5,8 +5,10 @@
 // run CI and contract tests in order to cancel, use certified, etc.
 
 // standard setup, present in every test
-const test = require("tape");
-const fs = require("fs");
+const tape = require("tape");
+const _test = require("tape-promise").default;
+const test = _test(tape);
+var fs = require("fs");
```
hilary, line 11: I believe you mean `const` :grin: Darn autocomplete!

hilary (changes requested):

package.json
```diff
@@ -26,7 +26,7 @@
         "slackchannel": "#billing-errors",
         "resources": {}
       },
-      "pmapi": {
+      "pmApi": {
```
hilary, line 29: It just occurred to me: you need to change line 20 of `.github/workflows/monitor.yml`.

I think we should have separate jobs in the monitor.yml file for each goalie, that way the goalie can drill down to the specific job rather than skimming through the collected tests.

hilary (changes requested):
Gah, just realized a change we need. While I'm at it, I have a suggestion.

package.json
sudoku-lord: I think that might be best to do in another PR (I can make the change right after this is merged).

hilary (approved):
:shipit:

tests/addresses_test.js
```diff
@@ -1,7 +1,9 @@
 "use strict";
 
 // standard setup, present in every test
-const test = require("tape");
+const tape = require("tape");
+const _test = require("tape-promise").default;
```
tlhunter, line 5: If `_test` is never used then this is perfectly valid:

```js
const test = require("tape-promise").default(tape);
```

Admittedly it's not pretty...

PR #85, opened 2021-04-29
convert self-mailers to use form factor models and other improvements

hilary:
As I got into checks, I realized that I wanted to reuse some of what @zacleids did with Self Mailers, so I decided to do this ticket first. Then, as I got into it, I realized that there was some cleanup that was overdue, so included that to help make things clearer to contributors.

https://lobsters.atlassian.net/browse/DXP-143

It may be easiest to review this PR as a series of small commits, which is what it is! :smiley_cat:


sudoku-lord (approved):

zacleids (approved):
lgtm

PR #87, opened 2021-05-04
added script to run single test file

sudoku-lord:


sudoku-lord:
I don't think the guardrails comment is correct at this point. I've added Joi validation for the command line input, so the string being passed into `exec` is safe.

run_one.js
```diff
@@ -0,0 +1,27 @@
+const Joi = require("joi");
+const pkg = require("../../package.json");
```
hilary, line 2: When I run `npm run one_test` I get an error because it can't find `'../../package.json'`, which makes sense because they are at the same level... 

A script run with `npm run ...` runs in the root directory of the package.

PR #94, opened 2021-05-18
Error enum

sudoku-lord:
I've added the codes from the docs to the error model, after a lot of headaches.

I know the description is repeated across the error code types. If there's a better way to do this then I would love to learn that.

rejoicing


hilary (changes requested):
Requested changes as discussed in slack...

package.json
```diff
@@ -55,7 +55,7 @@
     "postman": "npm run bundle && openapi2postmanv2 --pretty --spec dist/$npm_package_config_bundlefile --output dist/$npm_package_config_postmanfile",
     "pretty": "prettier --write .",
     "pretty:check": "prettier --check .",
-    "proxy": "prism proxy --errors $npm_package_config_specfile https://api.lob.com/v1",
+    "proxy": "prism proxy $npm_package_config_specfile https://api.lob.com/v1",
```
ajorczak-lob, line 58: Is the `errors` flag unnecessary now?

sudoku-lord: Unfortunately, the `errors` flag messes up the response. I'm actually going to convert this to a draft and wait to consult with Hilary, because the fix which was previously implemented doesn't seem to cover this.

shared/models/error.yml
```diff
@@ -1,10 +1,8 @@
 type: object
-
```
hilary, line 2: Please leave the whitespace in for readability.

shared/models/error.yml
```diff
@@ -1,10 +1,8 @@
...
 description: >-
   Lob uses RESTful HTTP response codes to indicate success or failure of an
   API request. In general, 2xx indicates success, 4xx indicate an input error,
   and 5xx indicates an error on Lob's end.
-
```
hilary, line 7: As above, leave whitespace in for readability.

shared/models/errors/advanced.yml
```diff
@@ -0,0 +1,33 @@
...
+  - invalid_international_feature
+  - invalid_perforation_return_envelope
+  - invalid_template_html
+  - merge_variable_required
+  - merge_variable_whitespace
+  - payment_method_unverified
+  - pdf_encrypted
+  - special_characters_restricted
+  - unembedded_fonts
+description: >
```
hilary, line 32: Where are the line item descriptions for the individual enum error values?
For now, let's include the status code in the enum description, so that the resulting documentation looks roughly the same as it does in the current docs.

hilary (changes requested):

shared/models/errors/advanced.yml
```diff
@@ -0,0 +1,62 @@
...
```
hilary, line 32: yaml folding, not for the faint of heart! I believe the operator you want here is `|`

shared/models/errors/authentication.yml
```diff
@@ -0,0 +1,16 @@
+type: string
+enum:
+  - email_required
+  - invalid_api_key
+  - publishable_key_not_allowed
+  - rate_limit_exceeded
+  - unauthorized
+  - unauthorized_token
+description: >
```
hilary, line 9: as above

shared/models/errors/generic.yml
```diff
@@ -0,0 +1,26 @@
...
+  - feature_limit_reached
+  - internal_server_error
+  - invalid
+  - not_deletable
+  - not_found
+  - request_timeout
+  - service_unavailable
+  - unrecognised_endpoint
+  - unsupported_lob_version
+description: >
```
hilary, line 14: as above

shared/models/errors/advanced.yml
```diff
@@ -0,0 +1,62 @@
...
+  A pre-defined string identifying the error category
```
hilary, line 33: Could we please remove the "Error ", as in "Error 422:"?

Ideally, we would reproduce the tables that are present in the current documentation in the descriptions. CommonMark doesn't include tables, but it does have tabs. You could use tabs to create a table-like appearance.

Alternatively, you could check to see what flavor of markdown Redocly supports. CommonMark turns out to be the required minimum by OAS v3. Supersets of CommonMark are allowed. GFM is the most common superset and it does support tables. https://github.github.com/gfm/#tabs

hilary (changes requested):

sudoku-lord:
lol I can fix this up, easy. to-do when I get back

sudoku-lord:
Rewriting this.

PR #98, opened 2021-05-20
Add slack alerts

sudoku-lord:
Automatically messages the appropriate channels in the case of contract test failures.

This is a sample error which could be surfaced in one of the goalie channels, due to a contract test failure:

```
FAILED CONTRACT TEST:      
 at /github/workspace/tests/zip_lookups_test.js:43:5
 
EXPECTED:  200
ACTUAL:      422
```

The original message contained a stacktrace, but I removed that because most of the stacktrace (save for the failed test line included in the current message) shows the parts in the `tape` module which errored (and those will stay the same regardless of which test fails).

The particular error message, in case of a contract test failure triggered by a mistake in production on our end, will always be "should be strictly equal", because we'd be surfacing a status code that doesn't match the one we expect in the tests. The test summary (e.g. "use an incorrectly formatted zip code") doesn't change from run to run, so that too can be excluded. Therefore, so that this doesn't clutter up any goalie channels, only the most important information––which contract test failed, the received status code, and the expected status code––will be surfaced in the message.

Assigning @jekatlob as well as Hilary, because it'll be good to have feedback from somebody who may be using these Slack alerts.

### DIFF
--- a/actions/contract_tests/goalieruns.js
+++ b/actions/contract_tests/goalieruns.js
@@ -42,7 +42,7 @@ module.exports.runTests = async function runTests() {
                 stdout.indexOf("      at:")
               );
               const block = [
-                `_FAILED TEST_:      ${testMessage}`,
+                `_FAILED CONTRACT TEST_:      ${testMessage}`,
                 `_EXPECTED_:  ${expectedCode}`,
                 `_ACTUAL_:      ${actualCode}`,
               ];
--- a/actions/contract_tests/goalieruns.js
+++ b/actions/contract_tests/goalieruns.js
@@ -1,19 +1,68 @@
+const Joi = require("joi");
 const pkg = require("../../package.json");
 const util = require("util");
 const exec = util.promisify(require("child_process").exec);
+const { WebClient } = require("@slack/web-api");
 
-for (let arg of process.argv.slice(2)) {
-  test_set = pkg.config.goalieMappings[arg].resources;
-  for (resource_name in test_set) {
-    test = test_set[resource_name];
-    test_command = 'multi-tape "' + test + '" | tap-spec';
-    exec(test_command, function (err, stdout, stderr) {
-      if (err) {
-        // actually: send this to the right slack channel
-        console.error(err);
-        return err.code;
+web = new WebClient(process.env.SLACK_TOKEN);
+
+module.exports.runTests = async function runTests() {
+  const validator = Joi.string()
+    .regex(/^[a-zA-Z]*$/)
+    .required();
+
+  for (let arg of process.argv.slice(2)) {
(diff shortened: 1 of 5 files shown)


actions/contract_tests/goalieruns.js
```diff
@@ -1,19 +1,49 @@
...
+      validated_arg = await Joi.compile(validator).validateAsync(arg);
+      test_set = pkg.config.goalieMappings[validated_arg].resources;
+      for (resource_name in test_set) {
+        test = test_set[resource_name];
+        test_command = 'multi-tape "' + test + '" | tap-spec';
+        exec(test_command, async function (err, stdout, stderr) {
+          if (err) {
+            try {
+              // send this to the right slack channel
+              const result = await web.chat.postMessage({
```
jekatlob, line 25: Two bits of feedback/requests:
1) can we enclose the message in a code block rather than posting it as plain text? Or format it the way Sentry and Datadog format their error messages
2) seeing the entire stack trace in the channel seems like it might get noisy and cumbersome. are the results stored somewhere else we could link to instead of posting the full error?

sudoku-lord: 1. I can figure out how to do that!
2. The stack traces for these errors are always pretty short (same fixed length as in the example channels), so it likely won't get too noisy unless there are a TON of tests failing. What do you mean by stored somewhere else––as in, where would they be stored?

hilary: They're in the monitoring run. 

I'm of two minds about your suggestion. Contract tests should not fail in production. When they do, it's a major crisis that should be immediately addressed. All the info you need should be right there without needing to go elsewhere.

Otoh, I can see a situation in which an existing crisis would cause the monitor to fail and you wouldn't want the monitor to swamp the channel. In that situation, you can always disable the monitor temporarily.

actions/contract_tests/goalieruns.js
```diff
@@ -1,19 +1,68 @@
...
+      for (resource_name in test_set) {
+        test = test_set[resource_name];
+        test_command = 'multi-tape "' + test + '" | tap-spec';
+        exec(test_command, async function (err, stdout, stderr) {
+          if (err) {
+            try {
+              // send this to the right slack channel
+              // the start and end indices, as well as the length, are constant because the stacktrace always
+              // includes the same spots in the tape module (only the failing test itself may change)
+              const startIndex = stdout.indexOf(
```
sudoku-lord, line 27: The start and end indices will actually be calculated from identical text every time (neither of these strings change regardless of the error being surfaced). That's why they're hard-coded in (I also added a comment to this effect).

senechko (approved):

PR #103, opened 2021-06-03
fix alerts

sudoku-lord:
Because of the Slack alerts malfunctioning on Saturday morning, I rewrote the script to account for a few different things:

## Indexing

This script still string indexes into the `tape` output in order to find the correct error slice to send as a message. Indexing itself shouldn't be an issue or take too long, since the `tape` output should be about the same length each time & the time performance of the runs does not appear to be affected (images below; the times naturally vary by 5-10 seconds regardless of the changes). The previous indexing system I used assumed that any errors would stem from the tests themselves, and therefore did not account for things like timeouts (which is what occurred on Saturday). The starting and ending points I've chosen for this update are present in every error trace from the `tape` output, regardless of what throws the error. I would like to figure out how to test this, so I've marked this PR as "Do Not Merge" until I add some testing in.

## Promises

Previously, an alert would be sent in the middle of the tests whenever a failure occurred. This has some issues: if Slack's Web API goes down and the `postMessage` function times out, the rest of the contract tests could be thrown off. That's why I've pushed each failure into an array and sent them to the appropriate channel all together after the full run.

## Contract Test Failure

Even when tests would fail, they were still marked with a green check by the action. This is because the script was not setting a failure using `core` in case of failed tests. I have added that in.

## Images

Time taken by tests which do not have slack alerts (and therefore do not perform any string indexing):

Screen Shot 2021-06-02 at 6 51 46 PM

Time taken by tests with slack alerts (and failures):

Screen Shot 2021-06-02 at 6 56 35 PM

Example messages (they are all grouped together regardless of goalie because in my testing repo I hardcoded the Slack channel to be a private one I created for this purpose):

Screen Shot 2021-06-17 at 12 05 37 PM


actions/contract_tests/goalieruns.js
```diff
@@ -10,62 +11,53 @@ module.exports.runTests = async function runTests() {
...
-    }
+    }).then(async function () {
+      if (failures.length > 0) {
+        let errorMessage = "";
+        if (failures.length > 1) {
+          errorMessage = `There were ${failures.length} failures in the contract tests:`;
+        } else {
+          errorMessage = `There was 1 failure in the contract tests:`;
+        }
+        try {
```
sudoku-lord, line 45: If this times out, the catch will capture the error Slack throws: https://slack.dev/java-slack-sdk/guides/web-api-basics

hilary:
It's a clever approach, but the output we get isn't fixed, so why process after the fact? Why not make tape produce the output we want in the first place? Instead of passing the output through `tap-spec`, let's write our own reporter which returns exactly what we want. That would have the added bonus of getting rid of a bunch of the vulnerabilities that `npm audit` is unhappy about, as `tap-spec` hasn't been updated in eons.

actions/contract_tests/goalieruns.js
```diff
@@ -10,62 +11,53 @@ module.exports.runTests = async function runTests() {
...
-          //     console.error(error);
-          //     return error.code;
-          //   }
-          // }
           if (err) {
-            console.log(err);
-            return err.code;
+            const startIndex = stdout.indexOf("✖");
+            const endIndex = stdout.indexOf("stack:");
+            failures.push(stdout.slice(startIndex, endIndex));
```
lbirdeau, line 27: This one section could be more defensive. Can we assume that the format can be guaranteed? Perhaps a try/catch or a defensive check in the preceding if statement would help.

actions/contract_tests/goalieruns.js
```diff
@@ -10,62 +11,53 @@ module.exports.runTests = async function runTests() {
...
+          }
+          if (++count == test_set.length) {
```
lbirdeau, line 29: The semaphore pattern works here, but the code might not be as readable for future maintainers trying to understand the strategy behind it. Consider adding a comment, creating a single-purpose function (my preference) or making the variable name more descriptive.

actions/contract_tests/goalieruns.js
```diff
@@ -10,62 +11,53 @@ module.exports.runTests = async function runTests() {
...
+        if (failures.length > 1) {
+          errorMessage = `There were ${failures.length} failures in the contract tests:`;
+        } else {
+          errorMessage = `There was 1 failure in the contract tests:`;
+        }
+        try {
+          const result = await web.chat.postMessage({
+            channel: "more-tests",
+            text: `:sadpanda: ${errorMessage}
+  ${failures.join("\n")}`,
```
lbirdeau, line 49: I'm a big fan of Slack's Block Kit library. It might be useful in a future commit to leverage some formatting and layout capabilities to improve the readability of the generated error message. I found the error messages readable on desktop, but the notification alert that appeared on my iPhone could have been titled better by taking advantage of Block Kit's formatting options.

actions/contract_tests/goalieruns.js
sudoku-lord: I can add a comment!

actions/contract_tests/goalieruns.js
sudoku-lord: We can assume this format will be guaranteed, as it's present in every single error surfaced by `tape`. My original mistake was picking a part of the stacktrace which was guaranteed to be present in every error _originating from the tests_, but that didn't account for external errors (this one does).

sudoku-lord:
After working on the custom reporter for a few days (and creating a really nifty one too 😢), I came to the conclusion that string indexing into the tap-spec output is actually the most viable solution for the time being, given the constraints of Tape. I've tested this pretty extensively in my private repo, and ensured that it works for external errors like the ETIMEDOUT which surfaced a few weeks back. I also added a failure in case it does collect a giant stacktrace somehow, so that it won't spam any channels. I think it's ready!

Next step: rewriting eeeverything for the switch to [ava](https://github.com/avajs/ava).

hilary (approved):
:shipit:

PR #105, opened 2021-06-11
tests(addresses): added tests for international address creation

sudoku-lord:
The address contract test didn't have any international address coverage, so I added it.

### DIFF
--- a/tests/addresses_test.js
+++ b/tests/addresses_test.js
@@ -119,6 +119,54 @@ test("allows creation with just a company", async function (t) {
   t.equal(response.status, 200);
 });
 
+test("correctly creates an international address", async function (t) {
+  const params = {
+    description: "Harry - Office",
+    name: "Harry Zhang",
+    email: "harry@lob.com",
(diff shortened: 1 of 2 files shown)


tests/addresses_test.js
```diff
@@ -119,6 +119,54 @@ test("allows creation with just a company", async function (t) {
...
+    .then((client) =>
+      client.post(resource_endpoint, params, { headers: prism.authHeader })
+    );
+
+  await t.doesNotReject(Promise.resolve(response));
+  t.equal(response.status, 200);
+  t.equal(response.data.address_country, "CANADA");
+});
+
+test("errors on receiving incorrect international address input", async function (t) {
```
hilary, line 147: I don't think this test is testing incorrect international address input. The country code defaults to "US", therefore this test would be treated as an incorrect US address.

hilary (changes requested):

hilary (approved):
LGTM!

PR #109, opened 2021-06-21
feat(av-bulk-processor): add av bulk specs and tests

amaan-lob:


sudoku-lord (changes requested):
Other than the comment I left regarding the bulk int'l tag, it looks good!

sudoku-lord (approved):

PR #110, opened 2021-06-22
feat(testing): switch testing framework from "tape" to "ava"

sudoku-lord:
`ava` is much nicer than `tape`! It has inbuilt async support & power assert feature which can pinpoint exactly which clause is making a test fail, amongst some other cool features. The output is much cleaner, so no need to use a reporter like `tap-spec`. 

As part of this, I redid the slack alerts feature. This is how the alerts look, as per this PR:

Screen Shot 2021-06-22 at 10 30 39 AM

And for errors such as timeouts:

Screen Shot 2021-06-22 at 10 55 09 AM

Like the last version, these alerts will also error & not send if they're over a certain length, to prevent goalie channels from being spammed.

Regarding the testing of goalieruns: it's a bit of a complicated task. I've tested the code extensively in my own repository, and checked to make sure the new `ava` framework deals with things like test timeouts. The best strategy I have right now is:

- create a test file which contains errors and place it under the `tests/` folder, but name it in the pattern `*_testing.js` so that it won't be run when `npm test` is called
- create a new addition to `goalieMappings` in `package.json` which contains a dummy private Slack channel, the aforementioned `testing` file, and a "goalie" name like `slackTest`
- use that goalie name and call `npm run pergoalie -- [slack goalie]` within some kind of test? this is the step which hasn't really clicked for me.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,6 @@
     "openapi-to-postmanv2": "^2.6.0",
     "prettier": "^2.2.1",
     "redoc-cli": "^0.11.4",
-    "tap-spec": "^5.0.0",
     "uuid": "^8.3.2"
   },
   "engines": {
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "clean": "rm -rf dist",
     "lint": "spectral lint $npm_package_config_specfile",
     "mock": "prism mock $npm_package_config_--dynamic $npm_package_config_specfile",
-    "singleTest": "func() { tape ./tests/\"$1\"_test.js | tap-spec; }; func",
+    "singleTest": "func() { ava ./tests/\"$1\"_test.js; }; func",
     "pergoalie": "node actions/contract_tests/goalieruns.js",
     "postinstall": "husky install",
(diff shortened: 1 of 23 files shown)


package.json
```diff
@@ -75,20 +75,19 @@
...
     "@stoplight/spectral": "^5.9.1",
+    "ava": "^3.15.0",
     "btoa": "^1.2.1",
     "husky": "^5.0.9",
     "joi": "^17.4.0",
-    "multi-tape": "^1.6.1",
     "openapi-to-postmanv2": "^2.6.0",
     "prettier": "^2.2.1",
     "redoc-cli": "^0.11.4",
     "tap-spec": "^5.0.0",
```
hilary, line 90: Please remove `tap-spec too!`

hilary (changes requested):
One smol but tasty change :)

package.json
sudoku-lord: done

hilary (approved):
:shipit:

PR #111, opened 2021-06-22
Support discrimination between url parameters

hilary:
Adding anchors to url parameters to support engines in discriminating between them when they are polymorphic alternatives.

I also included a quick fix to an example to remove a warning. Polymorphic alternatives of different types (string vs object) are not recommended although legal. Putting a string example with encoding instructions for the object case was simply too much for one of our tools to handle and it was unhappy. Swapping out for an object example solves the problem.


hilary:
This PR has been rebased so many times that it no longer works with our CI/CD chain. Replaced with #115

PR #115, opened 2021-06-23
 Support discrimination between url parameters (replaces #111)

hilary:
#111 waited for review long enough that repeated rebases rendered it non-viable. This PR replaces it.

Adding anchors to url parameters to support engines in discriminating between them when they are polymorphic alternatives. There's still a second problem to solve in order to fully resolve the relevant models, but I wanted to get this fix in first, separately, as it is a clear improvement in any case. The hope was that a small PR would get a quick approval. Oh well.

I also included a quick fix to an example to remove a warning. Polymorphic alternatives of different types (string vs object) are not recommended although legal. Putting a string example with encoding instructions for the object case was simply too much for one of our tools to handle and it was unhappy. Swapping out for an object example solves the problem.


zacleids (approved):
lgtm

PR #116, opened 2021-06-23
Adding Intl Verification new header details

vmangwani:
Added a new header for Intl Verification, x-lang-output
Added a sample test to run the verification for the new header.


hilary (approved):
very nice!

PR #117, opened 2021-06-24
feat(testing): fixed up template tests

sudoku-lord:
The template tests could have benefited from use of `ava`'s `serial` mechanism, so I added it in.

What's cool is that you can add assertions into the hooks, so I moved some of the tests to the `serial.before` hooks in some files, in order to maintain the integrity of the async code. They're not counted as tests in the final count, but if they fail they'll definitely toss up some errors (pictured below–an openapi-alerts message triggered by a failed assertion within the `serial.before` hook in `checks_test.js`):

Screen Shot 2021-06-25 at 12 51 07 PM

### DIFF
--- a/dist/Lob-API-postman.json
+++ b/dist/Lob-API-postman.json
@@ -1,11 +1,11 @@
 {
     "item": [
         {
-            "id": "04e15dc9-9db4-4546-829f-103a11d76d89",
+            "id": "87d5aa0f-63bd-4484-a6ec-e4a37123bf01",
             "name": "addresses",
             "item": [
                 {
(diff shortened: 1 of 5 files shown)


hilary (changes requested):
We shouldn't replace correctly behaving promises with globals, that's going in the wrong direction. Let's look at this together tomorrow and see what we can come up with.

sudoku-lord:
Last night I got rid of the globals (I was wrong earlier–you CAN perform assertions in hooks).

sudoku-lord:
Got rid of the globals last night, and today I added `checks_test.js` into the melee.

hilary (approved):
Hot diggity dog! That is just so pretty and readable!

PR #118, opened 2021-06-24
fix polymorphic models

hilary:
Provide ability to discriminate between local files, remote urls and html template strings.

Yes, that's negative lookahead in that regex.

Closes [DXP-191](https://lobsters.atlassian.net/browse/DXP-191)


sudoku-lord (approved):

sudoku-lord:
Approved (except for the Prettier change!)

hilary:
> Approved (except for the Prettier change!)

gah, return of the cyclical prettier issue.

PR #119, opened 2021-06-26
feat(tests_with_full_payload): created success test w/ full payload f…

jho44:
…or resources w/o one

**Why** : Some of the resources aren't being tested with a full payload (i.e. with optional parameters) so our API description could be failing common cases
**What**: Created a test that takes a full payload and is supposed to resolve successfully for every resource that didn't have such a test
**Details**: 
 - Got full payload specification from either following references in resources' yamls or by looking at [the current documentation](docs.lob.com) or [the draft documentation](https://lob.github.io/lob-openapi/)
 - Also changed `geo_ip_sort` to type boolean in all the relevant files since that's more intuitive

**Notes**: Used Lob account with Developer privileges and Scheduled Mailing feature flag


sudoku-lord (changes requested):
Looks good, except for just one change!

sudoku-lord (approved):
Perfect!

PR #120, opened 2021-06-28
Easy overview

jho44:
## Jira
[ticket](https://lobsters.atlassian.net/browse/DXP-200)

## Description
- Added easy parts of Overview from docs.lob.com to the main yaml
- Set them all to have `x-traitTag: true`
  - understand that this means they're shown in the side menu (which I can't see with my previewer...), has no subitems (since these are just blocks of text), and render the description (which is all they consist of)
- Included same links as in docs.lob.com except for ones that reference docs.lob.com


Lob-API-public.yml
```diff
@@ -34,12 +56,58 @@ tags:
...
+
+      You should also anticipate that features in Beta may have functional or design limitations, 
+      and might change rapidly as we receive customer feedback and make improvements. In particular, 
+      new APIs in Beta may also go through more frequent versioning and version deprecation cycles 
+      than our more mature APIs.
+
+      If you are participating in a Beta program and want to provide feedback, please feel free to 
+      [contact us](https://lob.com/support#contact)!
+
+      **Current Beta Features**
```
hilary, line 82: Let's remove the list of current beta features, since it is wildly out of date. Instead, we should look for some way to mark features as beta dynamically in the menu, similarly to what is done with the current docs.

hilary (changes requested):
One quick change :)

hilary (approved):
Looks good to me! :shipit:

PR #121, opened 2021-06-29
fix(postman): got rid of most errors & sorted out files when running postman 

sudoku-lord:
The error which does show up when running `npm run postman` is due to having a polymorphic discriminator with variants of different types ––given the structure of the docs, it is unavoidable. It does not prevent the collection from being created.


Lob-API-public.yml
```diff
@@ -217,7 +217,7 @@ x-tagGroups:
 
   - name: US Verification API
     tags:
-      - US Verifications
+      - Address Verification
```
hilary, line 220: left over edit?

hilary (changes requested):

hilary (approved):
LGTM

PR #124, opened 2021-07-06
feat(one_ex_for_all): add example for each resource

jho44:
[ticket](https://lobsters.atlassian.net/browse/DXP-168)

Either added to existing example or created whole new one such that every request and response schema has a least one example with all of the fields that resource can have
Will either be at the topmost level or one $ref down.
Copy pasted example for each content type (application/json, application/x-www-form-urlencoded, etc.) since I didn't think using something like `externalValue` and creating a whole new file would be necessary...


sudoku-lord (changes requested):
Fix those merge conflicts and I'll take another look!

sudoku-lord:
I see a couple of issues:

- In `Retrieve address with given ID`, the example is showing up empty. You might need to shift it around a little given that `address` has undergone some drastic changes.
Screen Shot 2021-07-09 at 3 30 34 PM

- Where's the `from` field in the response examples for `Retrieve postcard with given ID` and for `List all postcards`?
Screen Shot 2021-07-09 at 3 32 27 PM
Screen Shot 2021-07-09 at 3 37 17 PM

- For all the `List` examples which accept international addresses in the `to` field, could you add an international address example?

sudoku-lord (changes requested):
See my comment.

jho44:
> I see a couple of issues:
> ...

1st one: done
2nd one: there will be some list examples where one entry in the array is incomplete because I just left the old one there and appended/prepended a complete one.
3rd one: on it🖖

sudoku-lord (approved):

PR #130, opened 2021-07-07
add other address placements

jho44:
[ticket](https://lobsters.atlassian.net/browse/DXP-217)

image
From this, see that we should add both missing address placement options in while noting that one of them will be deprecated soon.

### DIFF
--- a/resources/letters/models/letter_base.yml
+++ b/resources/letters/models/letter_base.yml
@@ -36,10 +36,14 @@ allOf:
         enum:
           - top_first_page
           - insert_blank_page
+          - bottom_first_page_center
+          - bottom_first_page
         description: |
           Specifies the location of the address information that will show through the double-window envelope. To see how this will impact your letter design, view our letter template.
             * top_first_page - (default) print address information at the top of your provided first page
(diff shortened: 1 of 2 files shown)


resources/letters/models/letter_base.yml
```diff
@@ -36,10 +36,14 @@ allOf:
         enum:
           - top_first_page
           - insert_blank_page
+          - bottom_first_page_center
+          - bottom_first_page
         description: |
           Specifies the location of the address information that will show through the double-window envelope. To see how this will impact your letter design, view our letter template.
             * top_first_page - (default) print address information at the top of your provided first page
             * insert_blank_page - insert a blank address page at the beginning of your file (you will be charged for the extra page)
+            * bottom_first_page_center - **(deprecation planned -  within a few months)** print address information at the bottom center of your provided first page
```
sudoku-lord, line 45: Minor quibble, but could you get rid of the " - " in the deprecation explanation?

sudoku-lord (changes requested):
Super minor change requested.

sudoku-lord (approved):

PR #133, opened 2021-07-08
bug(addresses): fix address schema displays

sudoku-lord:
In short:

- The discriminator is unnecessary: it adds complexity & prevents the schema from displaying properly. Getting rid of it doesn't actually negatively affect the functionality, but it allows the schema to display.

- Redocly doesn't understand how to deal with allOf at the beginning of the `address.yml` file in `shared/models`. This format totally messes it up:

```
allOf:
  - $ref: address_editable_base.yml
  - $ref: address_fields_us.yml
  - $ref: address_generated_base.yml
  - type: object
```

Given that, the DRY format doesn't quite work for this. I manually included all the fields in address_us and address_intl, and that solved the issues I was seeing in to and from for mailpieces, wherein I kept seeing Recursive instead of the expected schema. Note that [this isn't the first time](https://github.com/Redocly/redoc/issues/585) Redocly has had issues with false-positive recursion.

Weirdly, `allOf` at the beginning of a file DOES work elsewhere, but at this point I'm ready to throw up my hands and implement whatever works.

PLEASE go through this with a fine-tooth comb, including generating the docs and poring over them, to see if I broke anything 😓 I think there are some files I might be able to get rid of, if they're not being referenced at all, but I haven't consolidated that information yet.


jho44 (approved):
The generated redoc from this looks good. I saw that you fixed how `data` in the `responses` portions of "List all"...of the addresses and checks resources wasn't being rendered properly. I also checked that all the properties were there by comparing with the complete examples I'd made for every resource in another branch.

These are the only refs I saw you delete. From a quick finder search, address_generated_base isn't referenced by anyone so that can be safely deleted. (Tested locally and by checking `List all Checks > 200 Responses > data > to` renders properly.
image
image
But honestly I don't mind if you leave it; we might need it later, given how often we have to refactor things. So if you delete it, I'll reapprove.

SidneyAllen (approved):
Reviewed output of Redoc and address object displaying correctly.

PR #137, opened 2021-07-09
Fix limit param

SidneyAllen:
Query Param “Limit” - should be all lowercase


jho44 (changes requested):
Looks good except for the merge conflicts and Prettier test fails. Prettier is failing when you run it locally on everything because our local version of Prettier is different from the version used in a 3rd party Github Action. There's a [fix](https://github.com/lob/lob-openapi/pull/139) for that coming but for now, you'll need to undo the changes on tests/letters_test.js and tests/postcards_test.js and then commit with --no-verify 😥 annoying I know

PR #138, opened 2021-07-09
Fix enum display max number

SidneyAllen:


jho44 (changes requested):
Looks good except for merge conflicts and Prettier fails, the solution of which is mentioned [here](https://github.com/lob/lob-openapi/pull/137#pullrequestreview-703286867)

PR #145, opened 2021-07-12
idem query

jho44:
[ticket](https://lobsters.atlassian.net/browse/DXP-197)

Added idempotency query param to relevant endpoints. Deleted them from resources that don't have idempotency header/query in docs.lob.com (bank accounts, certificates, and templates).

### DIFF
--- a/resources/addresses/addresses.yml
+++ b/resources/addresses/addresses.yml
@@ -36,9 +36,6 @@ post:
   tags:
     - Addresses
 
-  parameters:
-    - $ref: "../../shared/parameters/idempotency.yml"
-
   requestBody:
     required: true
     content:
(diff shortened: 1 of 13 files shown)


SidneyAllen (commented):
@jho44 

One small fix needed. It looks like the idempotent header/query on the Address.  This should be removed.  Everything else looks good.

According to doc.lob.com

Lob supports idempotency for safely retrying POST requests to create postcards, self mailers, letters, and checks without accidentally creating duplicates.

jho44:
> @jho44
> 
> One small fix needed. It looks like the idempotent header/query on the Address. This should be removed. Everything else looks good.
> 
> According to doc.lob.com
> 
> Lob supports idempotency for safely retrying POST requests to create postcards, self mailers, letters, and checks without accidentally creating duplicates.

Good catch!

SidneyAllen (approved):

PR #146, opened 2021-07-12
bug(slack-alerts): collapsing verbose slack alerts

sudoku-lord:
Alright, so slack alerts pooped all over some of the error channels again. I'll make sure all the lob-openapi secrets work as intended ASAP. Here's a fix which should help:

- Gets rid of the fancy string indexing, since there are some external errors that doesn't work for
- Collapses the actual errors into a reply to the original message, so all you'll see is:
Screen Shot 2021-07-12 at 3 57 58 PM
- Changes the frequency of the contract tests: they're now run every 30 minutes.

It's been a little difficult for me to figure out how much information to include in the tests, since I don't have any kind of design I'm working off 😓 But I think this marries the potential usefulness of the verbose errors with a neat display, since the channel won't be clogged up by a giant set of errors each time GitHub decides to get the stomach flu.

### DIFF
--- a/actions/contract_tests/goalieruns.js
+++ b/actions/contract_tests/goalieruns.js
@@ -22,13 +22,8 @@ module.exports.runTests = async function runTests() {
         test_command = 'ava "' + test + '"';
         exec(test_command, async function (err, stdout, stderr) {
           if (err) {
-            const startIndex = stdout.indexOf("─");
-            const endIndex = stdout.lastIndexOf("─");
-            if (stdout.slice(startIndex, endIndex)) {
-              const formatted =
-                "```" + stdout.slice(startIndex, endIndex) + "```";
-              failures.push(formatted);
-            }
+            const formatted = "```" + stdout + "```";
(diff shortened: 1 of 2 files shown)


actions/contract_tests/goalieruns.js
```diff
@@ -22,13 +22,8 @@ module.exports.runTests = async function runTests() {
...
         exec(test_command, async function (err, stdout, stderr) {
           if (err) {
-            const startIndex = stdout.indexOf("─");
-            const endIndex = stdout.lastIndexOf("─");
-            if (stdout.slice(startIndex, endIndex)) {
-              const formatted =
-                "```" + stdout.slice(startIndex, endIndex) + "```";
-              failures.push(formatted);
-            }
+            const formatted = "```" + stdout + "```";
```
senechko, line 25: might want to escape or somehow remove any occurrences of ``` in the `stdout` string

senechko (approved):

PR #147, opened 2021-07-12
metadata properties spec

jho44:
[ticket](https://lobsters.atlassian.net/browse/DXP-105)

Tried to add to the metadata schema the specifications listed in the metadata description and the ticket 

How it looks like:
image
- notice that maxProperties isn't showing up, which is apparently an issue that has been reported but never resolved [here](https://github.com/Redocly/redoc/issues/1343)
- also notice that the property key is a string resembling a pattern -- this pattern has no real power since I don't think there's a way to put constraints on Object keys in OpenAPI...


SidneyAllen (commented):
@jho44  - I'm wondering if this property should be oneOf: type number or boolean.

I'm under the impression it's a JSON object with key-value pairs 

{"customer_id": "1234"}

Do you have any information about how the type would be anything other than a string?

SidneyAllen:
@jho44 - I reviewed the description of https://lobsters.atlassian.net/browse/DXP-105

It's my conclusion that the type for this property should be a string which contains up to 20 key-value pairs.  The values of the pairs can be string, boolean or number but the property it self is a JSON object stored as a string.

jho44:
> It's my conclusion that the type for this property should be a string which contains up to 20 key-value pairs. The values of the pairs can be string, boolean or number but the property it self is a JSON object stored as a string.

Sorry I just saw these; my Github notifs are botched. I think the descriptions mean that metadata's keys and values are encoded as strings when the content type is URL encoded. This is how a regular sample request with metadata looks
image
and this is info about URL encoded vs JSON content types
image.

From what I understand, URL encoding, by definition, takes a string and turns it into a "safer" string that can be transmitted across the internet. And if we're following a JSON format for the other content types, then the keys are strings while the values are numbers, bools, or strings. I don't think I can specify that keys must be strings in OAS though.

SidneyAllen (approved):
Going forward with oneOf where metadata[campaign] = "NYC2021" or "2021" or "TRUE"

PR #151, opened 2021-07-14
feat(ncoa): add NCOA & examples

sudoku-lord:
I've added NCOA test & live environments, with examples, to the docs. I looked into the change @SidneyAllen suggested (aka that the responses automatically be set to the `ncoa` ones when the links are clicked)––but Redocly doesn't provide a way to link to the examples at all, and there doesn't seem to be a way to change the displayed example automatically based on linking.


Lob-API-public.yml
```diff
@@ -125,6 +125,53 @@ tags:
       - [Java](https://github.com/lob/lob-java)
       - [Elixir](https://github.com/lob/lob-elixir)
 
+  - name: NCOA
+    x-traitTag: true
```
lbirdeau, line 129: I looked this up and see it is one of Redocly's custom extensions to the open-api spec. Based upon my initial review, I can see how it applies to the NCOA pattern here, but I want to understand it better long-term.

lbirdeau (approved):
LGTM

shared/models/address/address_us.yml
```diff
@@ -118,6 +104,20 @@ properties:
   date_modified:
     $ref: "../../../shared/attributes/timestamps.yml#/date_modified"
 
+  recipient_moved:
```
jho44, line 107: lol any particular reason you moved this?

resources/addresses/responses/all_addresses.yml
```diff
@@ -36,7 +36,6 @@ content:
           address_zip: 94107-1741
           address_country: UNITED STATES
           metadata: {}
-          recipient_moved: false
```
jho44, line 39: just pointing out that removing this will make this an "incomplete" payload

shared/models/address/address_us.yml
sudoku-lord: yeah it was in the wrong spot!

resources/addresses/responses/all_addresses.yml
sudoku-lord: hmm but the payload will look like that (without `recipient_moved`) for almost everyone. only if you actually have ncoa enabled will you even see that field, so I'm not sure we want to include it in the regular response.

Lob-API-public.yml
sudoku-lord: https://redoc.ly/docs/api-reference-docs/vendor-extensions/redoc-supported-extensions/ I think the best explanation of its use is that it renders the tag description? I should actually make the other NCOA tags traitTags too...

PR #161, opened 2021-07-22
add list params test to each relevant resource

jho44:
[ticket](https://lobsters.atlassian.net/browse/DXP-62)

Created a test for each resource that has a "List" operation. Some of them, like Addresses, will break if a certain number of entries is added but I counteract this possibility by:
1. Deleting any resources that are created during testing and
2. Making other tests less stringent and reliant on the exact number of records there are

### DIFF
--- a/tests/addresses_test.js
+++ b/tests/addresses_test.js
@@ -22,6 +22,107 @@ test("list addresses", async function (t) {
   t.assert(response.status === 200);
 });
 
+test("list addresses' parameters", async function (t) {
+  const list = async (body) => {
+    const response = await prism.setup().then((client) =>
+      client.get(`${resource_endpoint}?${body}`, {
+        headers: prism.authHeader,
+      })
+    );
+    t.assert(response.status === 200);
+    return response.data;
+  };
+
+  /* ################## LIMIT ################## */
+
+  const limit_body = new URLSearchParams({ limit: 8 });
+
+  const limit_response = list(limit_body.toString());
+
+  /* ################## BEFORE ################## */
+
+  const before_body = new URLSearchParams({
+    before:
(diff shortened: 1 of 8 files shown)


SidneyAllen (commented):
@jho44  -  4 tests failed 

Screen Shot 2021-07-22 at 11 12 10 AM
Screen Shot 2021-07-22 at 11 12 25 AM

jho44:
> @jho44 - 4 tests failed
> 
> Screen Shot 2021-07-22 at 11 12 10 AM
> Screen Shot 2021-07-22 at 11 12 25 AM

*sobbing* It's all so very fragile. Looking into it now...

tests/addresses_test.js
```diff
@@ -22,6 +22,107 @@ test("list addresses", async function (t) {
...
+  const metadata_response = list(metadata_body.toString());
+
+  /* ################## DATE_CREATED ################## */
+
+  let deepObj = {
+    date_created: { gte: "2021-07-19", lte: "2021-07-20T19:32:00.000Z" },
+  };
+
+  const date_body = new URLSearchParams({});
+  for (const key in deepObj) {
```
sudoku-lord, line 77: I might be missing something here, but why create an outer map when the inner map only has one key? You could just hardcode this without using nested loops (it might shave off a few milliseconds from the test run, but it's not super necessary, so it's your call).

sudoku-lord (approved):
Approved, but I do have a comment about the structure of a particular aspect of the test.

tests/addresses_test.js
jho44: Ahaha I copy pasted this from the URLencoded tests, where stuff like `metadata` could have multiple keys. Shaving now...

PR #166, opened 2021-07-23
add diff lang req samples to create addr

jho44:
[ticket](https://lobsters.atlassian.net/browse/DXP-247)

In my humble opinion, this test was a success. If you agree, I can make one big PR that adds req samples to all relevant endpoints


SidneyAllen (approved):
That looks very slick - please proceed @jho44

PR #170, opened 2021-07-26
feat(errors): added error classifications

sudoku-lord:
Broke the error model down based on the status codes listed out in the docs.

I also customised the errors for each resource, since the default error made no sense for most of them.


Lob-API-public.yml
```diff
@@ -130,9 +130,6 @@ tags:
       in the future by supplying the `send_date` parameter, that will override any
       cancellation window you may have for that product.
 
-  - name: Certificates
-    description: A certificate store to manage client identities.
-
```
jho44, line 135: Why was the certificates resource deleted? Is it because Verizon is the only one using it? Because by that logic, we should exclude batch endpoints as well, which I just finished the first draft for. But I don't see anything wrong with leaving it in, especially if it may pique the interest of other customers.

shared/models/errors/generic.yml
```diff
@@ -0,0 +1,26 @@
...
+  * `conflict` - 409/422: this operation would leave data in a conflicted state.
+  * `feature_limit_reached` - 403: the account has reached its resource limit and requires upgrading to add more.
+  * `internal_server_error` - 500: an error has occured on Lob's servers. Please try request again.
+  * `invalid` - 422: an invalid request was made. See error message for details.
+  * `not_deletable` - 422: an attempt was made to delete a resource, but the resource cannot be deleted.
+  * `not_found` - 404: the requested resource was not found.
+  * `request_timeout` - 408: the request took too long. Please try again.
+  * `service_unavailable` - 503: the Lob servers are temporarily unavailable. Please try agian.
+  * `unrecognised_endpoint` - 404: the requested endpoint doesn't exist.
+  * `unsupported_lob_version - 422: an unsupported Lob API version was requested.
```
jho44, line 26:
```suggestion
  * `unsupported_lob_version` - 422: an unsupported Lob API version was requested.
```

shared/models/errors/advanced.yml
```diff
@@ -0,0 +1,62 @@
...
+  - merge_variable_required
+  - merge_variable_whitespace
+  - payment_method_unverified
+  - pdf_encrypted
+  - special_characters_restricted
+  - unembedded_fonts
+description: |
+  A pre-defined string identifying Lob-specific errors
+  * `address_length_exceeds_limit` - 422: the sum of to.address_line1 and to.address_line2 cannot surpass 50 characters.
+  * `bank_account_already_verified` - 422: the bank account has already been verified.
```
jho44, line 35: If I'd known you were working on this, I'd point you to the Errors section that Sid put in recently...

Lob-API-public.yml
sudoku-lord: @jho44 It's a private endpoint––I confirmed with PMAPI. Therefore, it absolutely shouldn't be in Lob OpenAPI. 

Batch should also not be in Lob OpenAPI if it's not a public endpoint (and that includes endpoints hidden by account feature flags, like `intl_autocompletions`––I had to remove that from Lob OpenAPI a while back). I mentioned creating a private repo for internal endpoints, Lob AdminAPI: _that_ is where things like `certificates` go.

shared/models/errors/advanced.yml
sudoku-lord: Yeah, that was actually something I had intended to put in this PR way back when. This is separate, though, because this is the actual resource as opposed to just the documentation (I also wrote this months ago).

Lob-API-public.yml
SidneyAllen: To confirm @sudoku-lord - the private endpoints will find a home in a new repo called lob-adminiapi

SidneyAllen (approved):

jho44 (approved):

PR #178, opened 2021-07-30
added og js + scss + gulp configs

jho44:
[ticket](https://lobsters.atlassian.net/browse/DXP-265)


jho44:
Closing because apparently Gulp has too many vulnerabilities that Guardrails doesn't like. Will open another PR where I use either webpack or browserify + some other task runner to accomplish this.

PR #190, opened 2021-08-05
fix typos bothering bundling

jho44:
Noticed while running `npm run bundle` and `npm run postman` that some phat errors resulted.
image
image

Turns out there were some typos.


sudoku-lord (approved):
SNATCH those errors out of the dastardly grasp of the bundler

PR #192, opened 2021-08-06
ref navbar links + pricing ver update

jho44:
[ticket](https://lobsters.atlassian.net/browse/DXP-275)

Made navbar content at least more in sync with lob.com's navbar (particularly under the Pricing tab), though the core of the openapi docs's navbar is from docs.lob.com. (Don't know where the js causing the hover-to-pull-down-menu is coming from though)

Put the variables (e.g. {{LOB_ASSETS_URL}} back into templates.hbs so it'll be easier to transition to having those variables based on dev mode (dev vs prod vs staging) later on. And since all these vars can be found on the public site, I didn't put them into Github secrets -- just slapped them into package.json


SidneyAllen (commented):
@jho44 All the images are broken when I run and test locally

Screen Shot 2021-08-09 at 3 03 03 PM

jho44:
> @jho44 All the images are broken when I run and test locally
> 
> Screen Shot 2021-08-09 at 3 03 03 PM

I suspect this is because you haven't applied my redoc-cli patch to your installation of redoc-cli. If you run `npx patch-package` from the project root then run `npm run redoc` again, the proper docs should be generated.

SidneyAllen (approved):
Works now after installing the patch - thanks @jho44

PR #197, opened 2021-08-09
fixed up goalieruns silent failures

sudoku-lord:
Fixes silent contract test failures

## Checklist

- [ x] Up to date with `main`
- [ x] All the tests are passing
  - [ x] Prettier
  - [ x] Spectral Lint
  - [x ] Contract Tests
    - [ x] Delete all resources created in contract tests
- [ x] `npm run bundle` outputs nothing suspect
- [x ] `npm run postman` outputs nothing suspect

## Changes

Did a deep dive and redid the slack alerts so that unexpected errors would no longer fail silently. There was a gnarly little NodeJS nuance (a function which is supported in some other forms of JS is not supported in NodeJS) which was causing issues for a while, but because they were silent failures they slipped under the radar. Now, at long last, the Slack alerts have reached their Final Form.

Meh, if you see anything which seems suspect let me know.

Screen Shot 2021-08-09 at 1 48 00 PM
Fig 1. When there's a big external failure message & some errors from the tests themselves, the big message will just get translated to a generic error message rather than being sent as-is. Weird external failures sometimes happen 🤷🏽 

Screen Shot 2021-08-09 at 1 49 42 PM
Verbose error messages, which are useful for pinpointing the exact locations of failures, are now nested as replies to a brief summary so that they won't throw a huge amount of noise into error channels.

### DIFF
--- a/actions/contract_tests/goalieruns.js
+++ b/actions/contract_tests/goalieruns.js
@@ -22,7 +22,7 @@ module.exports.runTests = async function runTests() {
         test_command = 'ava "' + test + '"';
         exec(test_command, async function (err, stdout, stderr) {
           if (err) {
-            const formatted = "```" + stdout.replaceAll("```", " ") + "```";
+            const formatted = "```" + stdout.replace(/```/g, " ") + "```";
             failures.push(formatted);
           }
           // async: the code won't proceed to the next block until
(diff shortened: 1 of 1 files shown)


mercedesb (approved):
🚀 Gotta love a regex, this looks great to me!

jho44 (approved):

PR #202, opened 2021-08-13
feat(try-catch): updated all the tests to include catches for exceptions

sudoku-lord:
Fixes DXP-287

## Checklist

- [x] Up to date with `main`
  - [x] Prettier
  - [x] Spectral Lint
  - [x] Contract Tests
- [ ] `npm run bundle` outputs nothing suspect
- [ ] `npm run postman` outputs nothing suspect

_i've gone blind_

(the last two are unchecked bc I have to update my own node version)


jho44:
- got rid of `t.plan` since it wasn't really adding anything to our error msgs
- added `t.assert(false)` in `catch` so ava doesn't cry about having at least 1 assertion per test
- got rid of the `prismError.message` parts -- I didn't find them as informative as the autogenerated exception messages
- `console.error(prismError)` rather than `console.error(JSON.stringify...)` b/c the latter was just printing `{}`

sudoku-lord:
FINALLY found the best solution to the `{}` pretty-printing issue.

tests/addresses_test.js
```diff
@@ -13,24 +13,40 @@ const prism = new Prism(specFile, lobUri, process.env.LOB_API_TEST_TOKEN);
...
+      );
+      t.assert(response.status === 200);
+      return response.data;
+    } catch (prismError) {
+      if (Object.keys(prismError).length > 0) {
+        t.fail(JSON.stringify(prismError, null, 2));
+      } else {
+        t.fail(prismError.toString());
+      }
+      return prismError;
```
jho44, line 48: if the flow comes into this branch, it'd never make it to this `return` right? cuz the `t.fail` would stop the whole test?

jho44 (approved):
Beyond that one comment I made, everything seems good. And even if u leave that `return` in, it doesn't change the effect

tests/addresses_test.js
sudoku-lord: ah crap the return should be in the else block, thanks for catching that

PR #218, opened 2021-08-20
fixed expand button colour & redocly script

sudoku-lord:


package.json
```diff
@@ -58,7 +59,7 @@
...
+    "redoc": "redoc-cli bundle $npm_package_config_specfile -t actions/redoc/template.hbs --options=docs/theme.json --templateOptions=docs/template.json --title=\"Lob API documentation\" -o docs/index.html",
```
jho44, line 62: image
don't think separating it is working -sweat-

jho44: you might have to write a patch for this in `redoc-cli` cuz their `templateOptions` was broken to begin with

jho44 (changes requested):
separating out templateOptions don't work :( either calls for a patch or separating a diff way

PR #226, opened 2021-08-27
feat(us_verifications): Temporary removal of enhanced_deliverability due to setbacks 

Thexumaker:
- We haven't been able to decide what to do with this feature so we should remove it from the open-api


jho44 (changes requested):
Since you deleted `recent_success`, could you also delete that property in the corresponding endpoint's examples?

Thexumaker:
> Since you deleted `recent_success`, could you also delete that property in the corresponding endpoint's examples?

good catch!

jho44 (approved):

PR #237, opened 2021-09-15
feat(address-verification) Reverse geocode endpoint + zip lookup spelling fix

Thexumaker:
## Checklist

- [x] Up to date with `main`
  - [x] Prettier
  - [x] Spectral Lint
  - [x] Contract Tests
    - [?] Delete all resources created in contract tests

## Changes
- Adds in documentation for a reverse geocoding endpoint + contract test
- Fixes a couple visual issues with us zip lookups(Python and php labels were incorrect, lookups was misspelled)


resources/us_verifications/models/us_components.yml
```diff
@@ -110,14 +110,14 @@ properties:
     $ref: "../../../shared/attributes/address/state.yml"
 
   zip_code:
-    type: string
-    description: The 5-digit ZIP code
-    pattern: ^\d{5}$
+    allOf:
+      - description: The 5-digit ZIP code
+      - $ref: "../../../shared/attributes/address/zip_code.yml"
```
sudoku-lord, line 115: So this shared attribute actually includes the 9 digit zip code as a possibility (the pattern is `^\d{5}((-)?\d{4})?$`), but the descriptions here imply that the 5 digit zip and 4 digit add-on are separated.

I think it makes most sense to keep the original configuration intact for `zip_code` here, instead of referencing this attribute. Changing the attribute itself is out of the question, since it's referenced in multiple other models.

sudoku-lord (changes requested):
Looks good, pending one change.

sudoku-lord (approved):

PR #243, opened 2021-09-27
Updating 7 (out of 9) delete methods to remove inline 200 models from SDK generator

zolopane:

## Checklist

- [X] Up to date with `main`
- [X] All the tests are passing
  - [X] Prettier
  - [X] Spectral Lint
  - [X] Contract Tests
    - [X] Delete all resources created in contract tests
- [X] `npm run bundle` outputs nothing suspect
- [X] `npm run postman` outputs nothing suspect

## Changes
Addresses, Bank Accounts, Letter, Postcards, Self Mailers, Checks, and Templates have been updated. 
The following two new files have been created for each: 
1. ../shared/responses/<resource_type>_deleted.yml
2. ../models/<resource_type>_deletion.yml


shared/models/check_cancelation.yml
```diff
@@ -0,0 +1,11 @@
+description: >-
```
sudoku-lord, line 1: NITPICK: could you change the name of this file to `check_cancellation`? Cancelation isn't incorrect, but it's also very very rare compared to the double-L spelling.

shared/attributes/model_ids/gen_id.yml
```diff
@@ -0,0 +1,5 @@
+type: string
```
sudoku-lord, line 1: What is this file for? Where are we using generic IDs?

sudoku-lord (changes requested):
Instead of separate `_deletion` models for each resource, could you do something like this? Then you only need to create one model, and you can reference it for every mailpiece.

```
description: >-
  Lob uses RESTful HTTP response codes to indicate success or failure of an
  API request. In general, 2xx indicates success, 4xx indicate an input error,
  and 5xx indicates an error on Lob's end.
properties:
  id:
    oneOf:
      - $ref: "../../shared/attributes/model_ids/adr_id.yml"
      - $ref: "../../shared/attributes/model_ids/bank/bank_id.yml"
      - $ref: "../../resources/checks/attributes/chk_id.yml"
      - $ref: "../../resources/letters/attributes/ltr_id.yml"
      - $ref: "../../resources/postcards/attributes/psc_id.yml"
      - $ref: "../../resources/self_mailers/attributes/sfm_id.yml"
      - $ref: "../../resources/templates/attributes/template_html.yml"
  deleted:
    $ref: "../../shared/attributes/deleted.yml"
```

shared/models/template_deletion.yml
```diff
@@ -5,7 +5,8 @@ description: >-
```
sudoku-lord, line 5: Do you still need this file?

zolopane: I don't! I'm going to delete every iteration of this file across the resources

sudoku-lord (approved):

PR #245, opened 2021-09-28
End to end

sudoku-lord:
Add in a little bit of e2e testing.

## Checklist

- [x] Up to date with `main`
  - [x] Prettier
  - [x] Spectral Lint
  - [x] Contract Tests


cypress/integration/links_test.js
```diff
@@ -0,0 +1,94 @@
...
+  it("contains a link to the main Lob page", () => {
+    // makes sure the lob.com link goes to the right page
+    cy.get('*[class^="brand w-nav-brand"]')
+      .should("have.attr", "href")
+      .and("include", "https://lob.com");
+  });
+
+  it("contains the support link", () => {
+    // makes sure the support link goes to the right page
+    cy.get('*[class^="sc-gsWcmt gCgqqY"]')
```
bethqiang, line 15: Depending on your build system and what it's configured to do, these classes may be randomly generated each time - if that's the case, this may fail after the first time running the tests. 😅  if it's the same random classes each time (if that statement makes sense, lol), this should be fine though!

cypress/integration/links_test.js
```diff
@@ -0,0 +1,94 @@
...
+      .and("include", "https://support.lob.com");
+  });
+
+  it("contains the right email", () => {
+    // the email should be correct
+    cy.contains("lob-openapi@lob.com");
+  });
+
+  it("contains the licence link", () => {
+    // makes sure the support link goes to the right page
```
bethqiang, line 27: Ooops, copy pasta comment? 😬 🍝  (same with the next few tests too)

bethqiang (commented):
This looks like a great start! Have a couple of comments inline, and one general comment - it seems like you're using classes a lot to get elements, which is ... _fine_, but can be fairly brittle because if you make one styling tweak or if your build system is configured so that classes aren't the same from render to render, it could cause your tests to fail, even if _you_ haven't done anything. 😬 

I'd recommend instead trying to use HTML tags as much as possible ([Cypress's .get page](https://docs.cypress.io/api/commands/get#Arguments) has some good examples with `li`, `input`, `a`, etc.).

Additionally, I just learned about [the DOM testing library's Cypress version](https://testing-library.com/docs/cypress-testing-library/intro). We've been using the DOM testing library for Vue and it's been really great and fairly easy to use. This should allow you to accomplish the same sort of thing, where you can bypass using classes entirely. You can `findByText` and then there should be? a way to say `hasAttribute('href', 'https://lob.com')` (though I don't know what the syntax would be for the Cypress version off the top of my head).

cypress/integration/links_test.js
sudoku-lord: Thank you for bringing this up! Actually, the classes are generated by Redocly and they're static, which is why I figured it's safe enough to do this. I've verified that they don't change from build to build, so we should be good (even though it's a little frowned-upon in Cypress).

bethqiang: Got it. I would still recommend using HTML tags/other means whenever you can (again, because if someone somewhere decides they want a different class on a thing, your tests will break despite actual functionality not changing, whereas HTML tags should almost always remain the same unless you're actually changing what's on the screen), but again, using classes is ... fine. 🙃

bethqiang (approved):

cypress/integration/links_test.js
sudoku-lord: Based on our conversation, there are two main reasons which make me think that classes are the best option after all:

(1) The Redocly-generated class names are static and do not regenerate between deploys. Even changing the colour of an element is done by referencing the Redocly class name in our `template` Handlebars file.

(2) Not all of the links being tested are attached to elements which contain text, so the `get().contains()` chain doesn't always work. For example, there's one link which is associated with an image, so I can't use `contains()` to find that––but it does have a unique, static class name.

PR #247, opened 2021-10-05
Separate out deletion models

sudoku-lord:
The old `oneOf` format was nice to look at, but unfortunately the end result made it seem like any type of resource ID could be used to delete any type of resource (e.g. you could use `sfm_` to delete a postcard). This is, of course, inaccurate––so separate IDs it is. 

## Checklist

- [x] Up to date with `main`
  - [x] Prettier
  - [x] Spectral Lint
  - [x] Contract Tests


zolopane (approved):
LGTM

PR #251, opened 2021-10-07
Add ids

sudoku-lord:
I tested locally against the changes I made to `template.hbs`, and it works:
Screen Shot 2021-10-07 at 12 04 56 PM

## Checklist

- [x] Up to date with `main`
  - [x] Prettier
  - [x] Spectral Lint
  - [x] Contract Tests

### DIFF
--- a/dist/lob-api-bundled.yml
+++ b/dist/lob-api-bundled.yml
@@ -1,0 +1,14191 @@
+openapi: 3.0.3
+info:
+  title: Lob API (2020-02-11) | Specification
+  version: 1.2.0
+  description: >
+    The Lob API is organized around REST. Our API is designed to have
+    predictable, resource-oriented URLs and uses HTTP response codes to indicate
+    any API errors. <p> Looking for our [previous
+    documentation](https://lob.github.io/legacy-docs/)?
+  license:
+    name: MIT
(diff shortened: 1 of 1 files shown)


cypress/integration/links_test.js
```diff
@@ -60,36 +63,111 @@ describe("ensures introduction links and text are valid", () => {
...
+  });
+
+  it("contains the SDK link", () => {
+    // makes sure the SDK link goes to the right page
+    cy.contains("Lob SDK")
+      .should("have.attr", "href")
+      .and("include", "#tag/SDKs-and-Tools");
+  });
+
+  // sc-hKFxyN dmghQN
```
mwj8410, line 117: Comment for the future:

I don't love these . It would be nice if this code was abstracted out in a more readable way. Just off the top of my head, maybe create an object that maps a meaningful name to the value. So at the top of the file, or somewhere that makes sense if these tags are used across multiple files, declare an object like:

```
const PageSection_Enum = {
  Postcards: 'sc-hKFxyN dmghQN'
};
```

then use it like:
```
cy.get(`*[class^="${PageSection_Enum.Postcards}"]`)
```

Obviously adhere to our naming standards. My example code is consistent with a previous project that I was on

PR #254, opened 2021-10-13
fixed template a bit

sudoku-lord:
The contract test bullet point in the template wasn't quite correct, as we don't manually run contract tests––just running `npm test` should suffice, which is already covered in a previous bullet point.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,10 +4,9 @@ Fixes #\<INSERT ISSUE NUMBER HERE\>
 
 - [ ] Up to date with `main`
 - [ ] All the tests are passing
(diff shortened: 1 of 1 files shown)


rogue0137 (approved):

.github/pull_request_template.md
```diff
@@ -4,10 +4,9 @@ Fixes #\<INSERT ISSUE NUMBER HERE\>
 
 - [ ] Up to date with `main`
 - [ ] All the tests are passing
-  - [ ] Prettier
-  - [ ] Spectral Lint
-  - [ ] Contract Tests
```
mwj8410, line 9: Is the "Contract Tests" line going away?

PR #255, opened 2021-10-14
remove pipe from title and space for Bank Accounts tag

SidneyAllen:

## Checklist

- [X] Up to date with `main`
- [X] All the tests are passing
  - [X] Prettier
  - [X] Spectral Lint
  - [ X] Contract Tests
    - [ ] Delete all resources created in contract tests
- [X] `npm run bundle` outputs nothing suspect
- [X] `npm run postman` outputs nothing suspect

## Changes
Removed the pipe delimiter from the title
Removed spaces in Bank Accounts tag


sudoku-lord (changes requested):
Shouldn't the spaces be removed from all the instances of the Bank Accounts tag in `lob-api-public.yml` as well as the `bundled` file, not just from BankAccounts in `bundled`?

PR #277, opened 2021-10-25
Update Operation ID's

zolopane:
https://lobsters.atlassian.net/jira/software/projects/DXP/boards/73?selectedIssue=DXP-452

Updated all operations to lead with the resource instead of the name of the operation

bulkIntlVerifications became intlVerificationsBulk, so I've updated both the directory names and the relevant references
bulkUSVerifications became USVerificationsBulk


package.json
```diff
@@ -21,7 +21,7 @@
       "av": {
         "slackChannel": "#av-api-alerts",
         "resources": [
-          "tests/bulk_intl_verifications_test.js",
+          "tests/intl_verifications_bulk_test.js",
           "tests/bulk_us_verifications_test.js",
```
sudoku-lord, line 25: Why isn't this also changed to `us_verifications_bulk_test`?

sudoku-lord:
Shouldn't `resources/bank_accounts/bank_account_verify.yml` have an operation ID of `bank_account_verify`, to fit into the pattern?

sudoku-lord:
(1) You appear to have not renamed `tests/bulk_intl_verifications_test.js` to `tests/intl_verifications_bulk_test.js` as you intended.

(2) Why is the Bulk US resource not similarly renamed in `package.json`? Why is there a `bulk_us_verifications` folder _and_ a `us_verifications_bulk` folder both under `resources/`?

(3) You need to run `npm run bundle` and `npm run redoc` after fixing up everything, so that they don't contain wrong information. 

I suggest you go through these files with a fine-toothed comb and make sure there aren't pieces you missed, since they could cause issues down the line.

Also, as a tip, don't use quotes in branch names, because that makes it harder to check out to them (you have to escape the character).

sudoku-lord (changes requested):
Some mistakes and possibly incorrect renamings.

package.json
SidneyAllen: @zolopane and @sudoku-lord 

I'm seeing the changes to the Tag names removing whitespaces in the PR.  I don't think this should be merged.

mwj8410:
Closing in favor of #284

PR #281, opened 2021-10-29
feat: tracking for on-page navigation

mwj8410:
Fixes Jirat Issue [DXP-434](https://lobsters.atlassian.net/browse/DXP-434)

## Checklist

- [x] Up to date with `main`
- [x] Prettier
- [x] Spectral Lint

## Changes
Added an AnalyticsTracker class to the template that seeds index.html. On window load complete, this new class is initialized. There was no elegant way to subscribe to existing events or to subscribe to the hash change event. The exact reason for this was not illuminated, but my suspicion is that the library code managing on page navigation is short circuiting the events in some way.

The solution was to isolate the html elements that are interacted with as well as the scroll event and conditionally dispatch tracking http calls through gtag(). For scrolling, a rest period of 500 ms was arbitrarily selected as a meaningful delay.

I have verified that the outbound events are showing up (delayed) in Google analytics without further adjustment to that system.

The line `gtag('config', 'UA-41294851-1');` results in the page_view being recorded, but the way it is recorded does not include the hash tag. This line is required and I could not find a way in documentation to configure its behavior. However, knowing that a particular has was navigated to is important, so during the initialization operation, i added a conditional additional dispatch if there was a has value

Second, there is a brief period of time when the HTML content of the page is loaded, but the analytics initialization has not initialized when a user could begin to interact with the page. I'm not overly concerned about closing this window because it is brief, and all the options that occurred to me would have been brittle based on network performance as 3rd party libraries were loaded in.


zolopane (approved):
LGTM after Matthew walked me through the code

PR #282, opened 2021-10-29
fixed incorrect local file reference

sudoku-lord:
Gets rid of `nifty.htm` reference in example postcard create request, since that is impossible in JSON.

## Checklist

- [x] Up to date with `main`
- [x] Prettier
- [x] Spectral Lint


resources/postcards/postcards.yml
```diff
@@ -120,7 +120,7 @@ post:
             address_zip: "94107"
             address_country: "US"
           front: tmpl_a1234dddg
-          back: "./nifty.htm"
+          back: tmpl_a1234dddg
```
imWillX, line 123: nit: Could we use a template id here? ```tmpl_c94e83ca2cd5121```

imWillX (approved):
nice! thanks for fixing the back section.

PR #264, opened 2021-10-19
Tag name update

zolopane:
Removing white space. I did a million commits to get this one in, but I think this is correctly updated to main except for the white space

PR #285, opened 2021-11-08
fixes incorrect type for date filters

sudoku-lord:
## Checklist

- [x] Up to date with `main`
- [x] Prettier
- [x] Spectral Lint


mwj8410 (changes requested):
This PR contains #264 that should be removed from tracking. I previously used `git reset --hard e31c4a4` to remove that commit. I believe the hash should be the same here

PR #288, opened 2021-11-17
feat: rename operation ids to be entity first

mwj8410:
Fixes [DXP-452](https://lobsters.atlassian.net/browse/DXP-452)

## Checklist

- [x] Up to date with `main`
- [x] Prettier
- [x] Spectral Lint

## Changes
Renamed add operations Ids that fit the pattern of `<action>_<entity>` to `<entity>_action`. Please not that `us_verfication` and `intl_verification` entities were renamed to be `verification_<class>` as part of this. This is because the `bulk_<class>_verifications` operations were being affected to place "bulk" as the suffix/action. I did not change `us_zips` or `us_autocomplete` because they were not otherwise affected by the change and there is no corresponding international version.


mwj8410:
Closed because we are going to treat the entity renaming as separate from the operationId change

PR #295, opened 2021-12-07
fix: correct the response signature for return_envelope on Letter

mwj8410:
Fixes [DXP-568](https://lobsters.atlassian.net/browse/DXP-568)

## Checklist

- [x] Up to date with `main`
- [x] Prettier
- [x] Spectral Lint

## Changes

Correctly describes the return_letter propery


resources/letters/models/letter_base.yml
```diff
@@ -41,7 +41,20 @@ allOf:
       return_envelope:
         oneOf:
           - type: boolean
-          - type: string
+          - type: object
```
sudoku-lord, line 44: Could we make this inline object a `$ref` to `return_envelope.yml`? It's not a huge impact here, but it'll be nice when we eventually turn this into an SDK generation version.

resources/letters/models/letter_editable.yml
```diff
@@ -1,5 +1,4 @@
 allOf:
-  - $ref: "letter_base.yml"
```
sudoku-lord, line 2: Wait, why are we getting rid of this ref to `letter_base` if we're adding `return_envelope` to `letter_base`? Is `return_envelope` only present in letter responses, rather than editable letters? If so, why aren't we adding it to `letter` instead of `letter_base`, and keeping this reference to the latter here?

sudoku-lord (changes requested):
See provided comments.

resources/letters/models/letter_editable.yml
mwj8410: The reason is that we have 4 versions of Letter. 3 are read versions as described in Letter are certified, registered, and no_extra_service, which all extend letter base. The 4th version, editable, did extend from base, however, the behavior of return envelope differs between read and write operations.

When writing, it has two possible value types:
```yaml
oneOf:
  - type: string
  - type: boolean
```
whereas when reading, it has a different set of two possible value types:
```yaml
oneOf:
  - type: boolean
  - type: object
```

resources/letters/models/letter_base.yml
```diff
@@ -41,7 +41,8 @@ allOf:
       return_envelope:
         oneOf:
           - type: boolean
-          - type: string
+          - $ref: "./return_envelope.yml"
```
sudoku-lord, line 44:
```suggestion
          - $ref: "return_envelope.yml"
```

sudoku-lord (approved):
Approved save for one minor nitpick.

PR #297, opened 2021-12-13
feat(us_verifications): add valid_address field

amaan-lob:
## Checklist

- [x] Up to date with `main`
- [x] Prettier
- [x] Spectral Lint

## Changes

Added `valid_address` field to us_verifications response.


resources/us_verifications/models/us_verification.yml
```diff
@@ -42,6 +42,10 @@ properties:
       * `deliverable_missing_unit` – The address is deliverable to the building's default address but is missing secondary unit information. There is a chance the mail will not reach the intended recipient.
       * `undeliverable` – The address is not deliverable according to the USPS.
 
+  valid_address:
+    type: boolean
+    description: Indicates whether this address exists.
```
sudoku-lord, line 47: How is this different from the end result of address verification? I think this summary should be slightly more specific about the purpose of this field (you don't have to make it long, just a bit more detailed).

sudoku-lord (changes requested):
Minor change requested (see comment). Also, merge conflicts!

resources/us_verifications/models/us_verification.yml
amaan-lob: This is intended to show whether the address actually exists in the real world, not just that whether it is deliverable by the USPS or not. We intend to add non-USPS data sources to AV so this field will capture that additional data. I believe the description came from product, so let me check back with them if they want to change it.

PR #309, opened 2022-02-24
feat(us bulk verify): Add in missing code samples

Thexumaker:

## Checklist

- [ ] Up to date with `main`
- [ ] Prettier
- [ ] Spectral Lint
- [ ] `npm run bundle` outputs nothing suspect
- [ ] `npm run postman` outputs nothing suspect

## Changes
Adds in missing code samples for bulk verify
Couldn't get tests running locally


sudoku-lord (changes requested):
It looks like the `bulk_intl_verifications` resource is also missing code samples. Additionally, would it be a good idea to add 2 addresses to the list instead of 1, so it's clearer that this can take in multiple addresses?

Thexumaker:
> It looks like the `bulk_intl_verifications` resource is also missing code samples. Additionally, would it be a good idea to add 2 addresses to the list instead of 1, so it's clearer that this can take in multiple addresses?

My only thing about adding 2 addresses is that it can get a bit long and I think if the adress is a part of a list data structure it should be clear that you can put multiple addresses. Thoughts?

sudoku-lord:
The comment about 2 addresses being too long is valid. I think it's fine now.

sudoku-lord (approved):

PR #349, opened 2022-07-01
fixed return envelope value in letter response

sudoku-lord:
It looks like return envelope values in letter responses can either be `null` or a return envelope object. The previous version of the response included boolean, which is not an accepted response value for the return envelope field.

https://lobsters.atlassian.net/browse/DXP-1106


sudoku-lord:
No idea why Redocly is failing 😐

BennyKitchell:
@sudoku-lord  I have closed this PR because the linter GH action had run for 2 days, and it was the only way to stop it. You are welcome to re-open if needed.

PR #364, opened 2022-07-21
creatives+campaigns+uploads tests

jho44:
[ticket](https://lobsters.atlassian.net/browse/DXP-1166)

## Checklist

- [x] Up to date with `main`
- [x] Prettier
- [x] Spectral Lint

## Changes
Added contract tests for Campaigns, Uploads, and Campaigns resources.

Couldn't get the `Upload a File` endpoint to work with prism but it works with CURL.


tests/uploads_test.js
```diff
@@ -3,38 +3,215 @@
...
+    if (Object.keys(prismError).length > 0) {
+      t.fail(JSON.stringify(prismError, null, 2));
+    } else {
+      t.fail(prismError.toString());
+    }
+  }
+});
 
-// add any failure cases you need here
+test.after.always("delete campaign", async function (t) {
```
julielinter, line 202: **[non-blocking]:** We delete a campaign after the test is done. Should we delete the upload as well?

julielinter (approved):
Let a non-blocking comment but otherwise, LGTM! Nice work! 👍🏼

sudoku-lord (approved):

tests/uploads_test.js
jho44: Sorry I didn't see this sooner. The upload is deleted within the same test that creates it. That code can be found [here](https://github.com/lob/lob-openapi/blob/7999515617b79b1c2c5450457724b2571d706758/tests/uploads_test.js#L138-L143). Let me know if I missed something. Thanks for the review :)

PR #386, opened 2022-08-22
Pmapi 1708 add uploads to campaigns

julielinter:
Fixes [PMAPI-1708](https://lobsters.atlassian.net/browse/PMAPI-1708)

## Checklist

- [x] Up to date with `main`
- [x] Prettier
- [x] Spectral Lint

## Changes
Adds `uploads` property to Campaigns response
`uploads` is an array of a single Upload object that is associated with the Campaign (similar to `creatives` of a Campaign with only a single Upload object)


jho44:
@julielinter Sorry about the postman script not working. That will be resolved in [this PR](https://github.com/lob/lob-openapi/pull/384) but it has yet to be approved. So I think you can ignore that breaking :)

julielinter:
> @julielinter Sorry about the postman script not working. That will be resolved in [this PR](https://github.com/lob/lob-openapi/pull/384) but it has yet to be approved. So I think you can ignore that breaking :)

@jho44 thank you!

resources/campaigns/models/campaign.yml
```diff
@@ -30,6 +31,12 @@ allOf:
         description: An array of creatives that have been associated with this campaign.
         type: array
 
+      uploads:
+        description: A single-element array containing the upload object that is assocated with this campaign.
+        type: array
+        minItems: 0
+        maxItems: 1
```
BennyKitchell, line 38: Just want to make sure to call this out, I assume this will be bumped in the future. So we are using an array of one to keep this flexible?

BennyKitchell (approved):
Just had one clarifying question, nothing blocking 👍

resources/campaigns/models/campaign.yml
julielinter: @BennyKitchell in the future, if we allow multiple uploads/audience files per campaign, then yes, we'll remove the max. Right now, I decided to return the most recently created upload as it'll be the most relevant. Creatives is also a list of objects, but the same logic applies - the most recently created is the relevant one, and I might update creatives to match uploads to make things less confusing to users.

PR #387, opened 2022-08-23
fixed creatives description

sudoku-lord:
I got rid of some unnecessary requirements, and edited out the incorrect types for the front/back/file fields.


jho44 (approved):
LGTM. If you could run Prettier before merging, that'd be great.

resources/creatives/models/creatives.yml
```diff
@@ -14,10 +14,44 @@ postcardsCreative:
               $ref: "../../postcards/attributes/postcard_size.yml"
 
         front:
-          $ref: "../../postcards/attributes/front.yml"
```
jharrell, line 17: Are these files being referenced anywhere else? Maybe should be removed.

resources/creatives/models/creative_writable.yml
```diff
@@ -2,8 +2,6 @@ oneOf:
   - allOf:
       - title: Postcard Creative
         required:
-          - front
```
julielinter, line 5: I have concerns about removing the asset fields as required fields from the documentation as that might cause some users to create creatives without assets and then go through the campaign creation process and end up with blank mailpieces 😬

resources/creatives/models/creatives.yml
sudoku-lord: They are being referenced within the postcard model.

resources/creatives/models/creative_writable.yml
jharrell: Yeah I'd stick with making these required and move to making them required on the API side.

sudoku-lord: @julielinter Ah, got it. I wasn't sure how to deal with that, because they don't appear to be formally required, but I'll add the requirements back for now

jharrell (approved):

PR #410, opened 2022-11-04
DANG-1708: removed deprecated properties

siddharthpant92:
Part of https://lobsters.atlassian.net/browse/DANG-1708

## Checklist

- [x] Up to date with `main`
- [x] Prettier
- [x] Spectral Lint

## Changes
For  `uploads`, i've removed all mentioned of the `columnMapping` and `overwriteColumnMapping` properties since they are no longer valid


dist/lob-api-bundled.yml
```diff
@@ -21236,10 +21199,7 @@ paths:
             --header 'Content-Type: application/json' \
             -u YOUR_KEY_HERE: \
             --data-raw '{
-                "campaignId": "cmp_f33809b18b6f3ea8",
-                "columnMapping": {
-                    "name": "recipient"
-                }
```
erin-doyle, line 21242: Instead of just removing `columnMapping` should it be replaced with `mergeVariableColulmnMapping` or any of the other column mapping properties?

resources/uploads/attributes/merge_variable_column_mapping.yml
```diff
@@ -5,4 +5,4 @@ default: null
 example:
   name: recipient_name
   gift_code: code
-description: The mapping of column headers in your file to the merge variables present in your creative. See our <a href="https://help.lob.com/print-and-mail/building-a-mail-strategy/campaign-or-triggered-sends/campaign-audience-guide#step-3-map-merge-variable-data-if-applicable-7" target="_blank">Campaign Audience Guide</a> for additional details.
+description: The mapping of column headers in your file to the merge variables present in your creative. See our <a href="https://help.lob.com/print-and-mail/building-a-mail-strategy/campaign-or-triggered-sends/campaign-audience-guide#step-3-map-merge-variable-data-if-applicable-7" target="_blank">Campaign Audience Guide</a> for additional details. <br />If a merge variable has the same "name" as a "key" in the `requiredAddressColumnMapping` or `optionalAddressColumnMapping` objects, then they **CANNOT** have a different value in this object. If a different value is provided, then when the campaign is processing it will get overwritten with the mapped value present in the `requiredAddressColumnMapping` or `optionalAddressColumnMapping` objects.
```
erin-doyle, line 8: > If a merge variable has the same "name" as a "key" in the `requiredAddressColumnMapping` or `optionalAddressColumnMapping` objects

This wording is a bit confusing to me.  I wonder if you could just say:
> If a merge variable has the same "key" in the `requiredAddressColumnMapping` or `optionalAddressColumnMapping` objects

Rather than "name" as a "key".

erin-doyle: And also:
> have a different value in this object

Maybe that could be:
> have a different value for that key

dist/lob-api-bundled.yml
siddharthpant92: Those properties are already mentioned and [available publicly](https://docs.lob.com/#tag/Uploads/operation/uploads_list). So now i'm only removing all mentions of the deprecated properties.
Also, FWIW, this document is generated automatically with `npm bundle`

resources/uploads/attributes/merge_variable_column_mapping.yml
siddharthpant92: I was trying to say that if the merge variable is the same as a required/optional column like `name`, `address_line1` etc, then the merge variable cannot be mapped to a different value in `mergeVariableColumnMapping`. It has to be mapped to the same value as it is in `requiredAddressColumnMapping` or `optionalAddressColumnMapping`. 

So I was talking about the `merge_variable` **"name"** itself and the **"keys"** of the `requiredAddressColumnMapping` or `optionalAddressColumnMapping` objects.

But if that intention isn't clear enough, then i can definitely change it to your suggestion.

BennyKitchell (approved):

resources/uploads/attributes/merge_variable_column_mapping.yml
erin-doyle: I definitely didn't follow that by "name" you were saying "for instance if name is used as a key..."  I would recommend rewording it in some way, doesn't have to be what I suggested but I definitely don't think it's worded clearly enough.

PR #413, opened 2022-12-13
AV-3625 Update QR code documentation to add a pages as a key

vmangwani:
## Changes
Updates QR code documentation to add an extra key that we now accept called "pages" to stamp QR codes on specific pages.
This also modifies the letters and postcards create endpoint where qr_code object is accepted.

### DIFF
--- a/shared/models/qr_code.yml
+++ b/shared/models/qr_code.yml
@@ -36,3 +36,7 @@ properties:
   width:
     type: string
     description: The size(in inches) of the QR code. All QR codes are generated as a square.
+
+  pages:
+    type: string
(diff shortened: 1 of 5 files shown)


shared/models/qr_code.yml
```diff
@@ -36,3 +36,7 @@ properties:
   width:
     type: string
     description: The size(in inches) of the QR code. All QR codes are generated as a square.
+
+  pages:
+    type: string
+    description: Specify the pages where the QR code should be stamped in a comma separated format. For postcards, The values should be front and/or back. For letters, the values can be specific page numbers or page number ranges.
```
jho44, line 42:
```suggestion
    description: Specify the pages where the QR code should be stamped in a comma separated format. For postcards, the values should be front and/or back. For letters, the values can be specific page numbers or page number ranges.
```

jho44 (approved):
Outside of the small typo in the description, LGTM

PR #422, opened 2023-01-30
rm freddy feedback widget

senechko:
Removes Freddy Feedback Widget


BennyKitchell (approved):
Looks Good! (Also thank you!)

PR #432, opened 2023-04-20
AV-3752 add docs for qr code campaign creation

vmangwani:
Fixes text from a previous PR to add QR codes to campaigns


BennyKitchell (approved):
🚀

PR #441, opened 2023-06-26
Update change log portion to be removed

BennyKitchell:
Remove changelog since we have beamer to share new updates on the dashboard and this has been sunsetted


dist/lob-api-bundled.yml
```diff
@@ -3086,277 +3083,7 @@ tags:
 
 
       ### Changelog
```
juanfriss, line 3085: will this title still show? will the section by empty?

juanfriss (approved):

dist/lob-api-bundled.yml
BennyKitchell: Yup, it will show Changelog as the header and below it will be a link to the beamer changelog

PR #530, opened 2026-03-12
FD-164: Add Resource Proofs API documentation

NateWaldschmidt:
Fixes #FD-164

## Checklist

- [x] Up to date with `main`
- [x] Prettier
- [x] Spectral Lint

## Changes

Adds complete OpenAPI specification for the Resource Proofs resource, enabling users to create proofs for reviewing mailpiece creatives. The implementation includes:

**New Endpoints:**
- `POST /resource_proofs` - Create a new resource proof
- `GET /resource_proofs/{res_prf_id}` - Retrieve a resource proof
- `PATCH /resource_proofs/{res_prf_id}` - Update a resource proof (template_id only)

**Resource Types Supported:**
- Postcard
- Letter
- Self Mailer

**Schema Features:**
- Discriminated `oneOf` pattern for the three resource types
- Proper error handling with detailed issue entries
- Thumbnail support (small, medium, large)
- Complete model composition using `allOf` for reusability
- Consistent response formats aligned with existing Lob API patterns

**Files Added:**
- 16 new OpenAPI definition files under `resources/resource_proofs/`
  - Models (base, editable, updatable, resource-type-specific parameters)
  - Attributes (id, status, resource_type)
  - Responses (resource_proof, post_resource_proof)
  - Endpoint definitions

**Files Modified:**
- `lob-api-public.yml` - Added tag definition, tagGroup reference, and path definitions
- `docs/index.html` - Auto-generated from updated spec

- Response example in `resources/resource_proofs/responses/resource_proof.yml` is missing `resource_parameters` and `url` fields (see code review findings)
- Top-level `required` in `resource_proof_editable.yml` may need alignment with how other discriminated schemas handle this constraint
- Some unstaged changes needed (description replacement and example adjustments)

### DIFF
--- a/lob-api-public.yml
+++ b/lob-api-public.yml
@@ -1494,6 +1494,10 @@ tags:
         --------------------------7015ebe79c0a5f8c--
       ```
       <div class="back-to-top" ><a href="#" onclick="toTopLink()">back to top</a></div>
+  - name: Resource Proofs
+    description: |
+      The resource proofs endpoint allows you to create a final rendering of any template. This is best practice to ensure that you are visually validating your creative before any mail pieces use the template. The API provides endpoints for creating, updating, and retrieving template proofs.
+      <div class="back-to-top" ><a href="#" onclick="toTopLink()">back to top</a></div>
   - name: Reverse Geocode Lookups
     description: |
       Find a list of zip codes associated with a valid US location via latitude and longitude. <br> <div class="back-to-top" ><a href="#" onclick="toTopLink()">back to top</a></div>
@@ -2416,6 +2420,7 @@ x-tagGroups:
       - Booklets
       - Bank Accounts
       - Templates
+      - Resource Proofs
       - Template Versions
       - Template Design
       - Manage Mail
@@ -2563,6 +2568,10 @@ paths:
     $ref: resources/url_shortener/links_id.yml
   /links:
     $ref: resources/url_shortener/links.yml
+  /resource_proofs/{res_prf_id}:
+    $ref: resources/resource_proofs/resource_proof.yml
+  /resource_proofs:
+    $ref: resources/resource_proofs/resource_proofs.yml
   /self_mailers/{sfm_id}:
     $ref: resources/self_mailers/self_mailer.yml
   /self_mailers:
--- a/lob-api-public.yml
+++ b/lob-api-public.yml
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 info:
   title: Lob
-  version: 1.20.9
+  version: 1.21.0
   description: |
     The Lob API is organized around REST. Our API is designed to have predictable, resource-oriented URLs and uses HTTP response codes to indicate any API errors. <p>
   license:
--- a/resources/resource_proofs/models/resource_proof.yml
+++ b/resources/resource_proofs/models/resource_proof.yml
@@ -1,0 +1,27 @@
+allOf:
+  - $ref: "resource_proof_base.yml"
+
+  - type: object
+    required:
+      - id
+      - resource_type
+      - status
+      - thumbnails
+      - errors
+      - date_created
+      - date_modified
+      - object
+    properties:
+      id:
+        $ref: "../attributes/res_prf_id.yml"
+
+      date_created:
+        $ref: "../../../shared/attributes/timestamps.yml#/date_created"
+
+      date_modified:
+        $ref: "../../../shared/attributes/timestamps.yml#/date_modified"
+
+      object:
+        type: string
+        description: Value is resource type.
+        default: "resource_proof"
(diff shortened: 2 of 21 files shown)


KARSE22 (approved):

resources/resource_proofs/models/thumbnail.yml
KARSE22, line 1: I think in lob-api the resource_proof thumbnails are signed links. Could we use the thumbnail model in `shared/models/thumbnail.yml `? Seems like it types each URL field as a `signed_link` rather than `string`.

resources/resource_proofs/models/resource_proof.yml
```diff
@@ -0,0 +1,27 @@
...
+
+      date_created:
+        $ref: "../../../shared/attributes/timestamps.yml#/date_created"
+
+      date_modified:
+        $ref: "../../../shared/attributes/timestamps.yml#/date_modified"
+
+      object:
+        type: string
+        description: Value is resource type.
```
KARSE22, line 26: Should this be value is "resource_proof"? since the default is `resource_proof`?

resources/resource_proofs/models/resource_proof_editable.yml
```diff
@@ -0,0 +1,56 @@
...
+        enum:
+          - self_mailer
+        description: The type of resource to generate a proof for.
+      resource_parameters:
+        $ref: "self_mailer_resource_parameters.yml"
+
+discriminator:
+  propertyName: resource_type
+  mapping:
+    postcard: "#/oneOf/0"
```
KARSE22, line 54: I'm not 100% sure if this will work. Looking at the [docs](https://spec.openapis.org/oas/v3.0.3#discriminator-object) 

 These discriminator mapping values (`#/oneOf/0`, `#/oneOf/1`, `#/oneOf/2`) aren't valid OpenAPI 3.0 $ref paths. When it gets bundled, it looks like they compile to schemas like this ('0', '1', '2'), and I don't believe those are valid, since they need to be `$ref`'s? I think we'd have to give each `oneOf` variant its own named schema file.

KARSE22:
not a blocker: wondering if we'd like to add tests for resource proofs? It seems like some of the other resources have contract tests. But this might be harder since the `status` field is low key async and we might need to poll? But maybe we could just verify the initial `processing` state?

lob-api-public.yml
```diff
@@ -1494,6 +1494,10 @@ tags:
         --------------------------7015ebe79c0a5f8c--
       ```
       <div class="back-to-top" ><a href="#" onclick="toTopLink()">back to top</a></div>
+  - name: Resource Proofs
+    description: |
+      The resource proofs endpoint allows you to create a final rendering of any template. This is best practice to ensure that you are visually validating your creative before any mail pieces use the template. The API provides endpoints for creating, updating, and retrieving template proofs.
```
KARSE22, line 1499: Wondering if we should say any template here, since it's only available for certain template types?

resources/resource_proofs/models/resource_proof_base.yml
KARSE22, line 1: Should we add `resouce_parameters` since we return it in the response?

lob-api-public.yml
NateWaldschmidt: This was copy from product

resources/resource_proofs/models/resource_proof.yml
NateWaldschmidt: I copied this from some other resources in these docs. Its a weird pattern that we include these though

resources/resource_proofs/models/resource_proof_base.yml
NateWaldschmidt: I left them off for now since they aren't needed to create resource proofs from an API perspective

resources/resource_proofs/models/resource_proof_editable.yml
NateWaldschmidt: Screenshot 2026-03-16 at 11 17 15 AM

This is what I see when testing which is what I want. Does this look like you would expect?

resources/resource_proofs/models/thumbnail.yml
NateWaldschmidt: Updated

resources/resource_proofs/models/resource_proof.yml
KARSE22: Ahh okay! I was just scanning it and thought it was odd. sounds good!

resources/resource_proofs/models/resource_proof_editable.yml
KARSE22: This looks good, thanks for the screenshot! I didn't test the UI, I was just going off what I noticed in the `lob-api-bundled.yml` and wasn't sure if it would cause any issues.

vmangwani (approved):

